PR #80, opened 2013-04-06
Update pl.yml

wkulesza:
Modified and extended pl translation


assertchris:
Thanks for this but it would be better to add translation stuff on [Transifex](https://www.transifex.com/projects/p/silverstripe-blog). This file will just get overridden the next time we update from Transifex...

PR #99, opened 2013-08-13
BUG: Fixes #84 for tag 0.6.2 so blog can be loaded via composer

### LINKED ISSUE #84: Composer dependency mismatch on 0.6.2
0.6.2 can't be installed on SS 3.0.x using Composer as it requires silverstripe/widgets dev-master which in turn requires silverstripe/cms >= 3.1

kmayo-ss:
I would like this pull request to be created as a new tag 0.6.3 this is to resolve the following issues

https://github.com/silverstripe/silverstripe-blog/issues/84

https://github.com/silverstripe/silverstripe-widgets/issues/49

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,6 +17,6 @@
 	"require":
 	{
 		"silverstripe/cms": "3.*",
-		"silverstripe/widgets": "dev-master"
+		"silverstripe/widgets": "*"
 	}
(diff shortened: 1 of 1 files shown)


tractorcow:
Widgets isn't a dependency anymore

PR #126, opened 2014-08-05
Add extension hooks to blog entry form

mlewis-everley:
If we create more complex blog pages (with fields beyond the fields provided by this module) then we cannot edit this data using the "BlogEntryForm".

Adding extension hooks on both the BlogEntryForm and the postblog methods would allow developers to product more complex forms without having extend BlogHolder (and add additional, possibly unnecessary Model classes to the SiteTree).

Just a thought :-)

Mo

### DIFF
--- a/code/BlogEntry.php
+++ b/code/BlogEntry.php
@@ -44,19 +44,19 @@ class BlogEntry extends Page {
 	static $allow_wysiwyg_editing = true;
 	
 	/**
-	 * Overload so that the default date is today.
+	 * Overload so that the default date is today and the default author is the logged in Member.
 	 */
 	public function populateDefaults(){
 		parent::populateDefaults();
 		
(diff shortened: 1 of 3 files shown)


code/BlogEntry.php
```diff
@@ -44,19 +44,19 @@ class BlogEntry extends Page {
...
 	
 	/**
-	 * Overload so that the default date is today.
+	 * Overload so that the default date is today and the default author is the logged in Member.
 	 */
 	public function populateDefaults(){
 		parent::populateDefaults();
 		
 		$this->setField('Date', date('Y-m-d H:i:s', strtotime('now')));
+		$this->setField('Author', Member::currentUser() ? Member::currentUser()->getName() : '');
```
tractorcow: Make sure you call Security::database_is_ready() here before trying to access the current user.

dhensby:
Fixed by #153

PR #135, opened 2014-10-24
Phillprice blog entry patch

dhensby:
extending on #134


dhensby:
I'm going to close this if as #140 is a better fix and this is trying to do too much.

I'll look at implementing `SS_Datetime::now()` seperately.

PR #140, opened 2014-12-08
BUG Fix issue with populateDefault breaking during dev/build

### LINKED ISSUE #139: populateDefaults() causes MySQL error during build
As per https://github.com/silverstripe/silverstripe-framework/issues/3660 the populateDefaults function can cause a build error if the Member database schema is modified as it tries to access the logged in Member before the database is built. This will probably be an issue for lots of people when upgrading from SS3.1.6 or earlier due to new fields on Member.

Solution seems to be to wrap the populateDefaults function in a `if (Security::database_is_ready())` check.

tractorcow:
Fixes #139

### DIFF
--- a/code/BlogEntry.php
+++ b/code/BlogEntry.php
@@ -50,7 +50,10 @@ class BlogEntry extends Page {
 		parent::populateDefaults();
 		
 		$this->setField('Date', date('Y-m-d H:i:s', strtotime('now')));
-		$this->setField('Author', Member::currentUser() ? Member::currentUser()->getName() : '');
+
+		if(Security::database_is_ready() && ($member = Member::currentUser())) {
+			$this->setField('Author', $member->getName());
+		}
 	}
 	
 	function getCMSFields() {


dhensby:
@tractorcow so what's causing these travis failures?

tractorcow:
https://github.com/composer/composer/pull/3480 needs merging :)

dhensby:
@tractorcow whilst that takes its time being pulled in, is there any work around so we can get passing tests?

tractorcow:
Then we have to make a call on whether or not we can dismiss the travis failures for the time being, in order to allow further pull requests to be merged. We've done it before in the past (e.g. during the big ORM rewrite debacle).

What does your better judgement tell you? :)

dhensby:
Well, I think this looks like a sensible fix, but we still have a problem despite that PR to composer being merged because there are too many satisfiable SS modules.

Perhaps we need to enforce that we only test against stable releases?

PR #142, opened 2015-01-12
BUG: AuthorID is still referenced

textagroup:
Sending an AuthorID URL parameter to a Blog controller will throw SQL errors because of missing the AND keyword and because the AuthorID column has been removed from BlogEntry


code/BlogTree.php
```diff
@@ -270,19 +270,29 @@ public function BlogEntries($limit = null) {
 		} else {
 			$filter = '';
 		}
+
+		// AuthorID has been removed from BlogEntry check for it incase it is an upgrade
+		$hasAuthorID = BlogEntry::get()->first()->hasDatabaseField('AuthorID');
```
dhensby: If there are no blog entries this will fail.

You want to use the singleton to check the DB fields:

```php

singleton('BlogEntry')->hasDatabaseField('AuthorID');

```

code/BlogTree.php
```diff
@@ -270,19 +270,29 @@ public function BlogEntries($limit = null) {
...
+			$filter .=  (empty($filter)) ?
+				" \"BlogEntry\".\"Author\" LIKE '". Convert::raw2sql($_GET['author']) . "'"
+				: " AND \"BlogEntry\".\"Author\" LIKE '". Convert::raw2sql($_GET['author']) . "'";
 		}
-		else if(isset($_GET['authorID'])) {
-			$filter .=  " \"BlogEntry\".\"AuthorID\" = '". Convert::raw2sql($_GET['authorID']). "'";
+		else if($hasAuthorID == true && isset($_GET['authorID'])) {
+			$filter .=  (empty($filter)) ?
+				" \"BlogEntry\".\"AuthorID\" = '%". Convert::raw2sql($_GET['authorID']). "%'"
+				: " AND \"BlogEntry\".\"AuthorID\" = '". Convert::raw2sql($_GET['authorID']). "'";
```
tractorcow: You could save a lot of repetitive code by putting your author condition in a single variable and then joining it into filter at the end.

code/BlogTree.php
tractorcow: Also, this will return true if you have an AuthorID on the Page class, which will break your below conditions. If you use the filter api it would work ok, but this seems to be using pretty old manual SQL.

To check if the BlogEntry class has an AuthorID (not inherited) you can use `DataObject::custom_database_fields('BlogEntry')` and check for the AuthorID and Author keys.

textagroup:
I can see there is a bit of movement on the blogs module recently and this pull request has gone stale.
Do you want me to close it and open another one (if it still needs to be fixed)?

tractorcow:
You can still open it on the 1.0 blog... sorry for breaking your pull request as the master branch now belongs to a whole new tree. ;)

PR #144, opened 2015-02-12
Compatibility with blog 1.0

tractorcow:
This change introduces:
- Default theme included in base templates (can be overridden in user templates) based on https://github.com/micmania1/simple_blog
- Inclusion of legacy migration classes for old BlogTree, BlogHolder and BlogEntry pages, as well as ArchiveWidget and TagCloudWidget.
- Migration task, which can be configured to be either manually run, or run automatically during dev/build (on by default)

### DIFF
--- a/code/compat/pages/BlogEntry.php
+++ b/code/compat/pages/BlogEntry.php
@@ -1,0 +1,72 @@
+<?php
+
+/**
+ * @deprecated since version 2.0
+ */
+class BlogEntry extends BlogPost implements MigratableObject {
+	
+	private static $hide_ancestor = 'BlogEntry';
+
+	private static $db = array(
+		"Date" => "SS_Datetime",
+		"Author" => "Text",
+		"Tags" => "Text"
+	);
+
+	/**
+	 * Safely split and parse all distinct tags assigned to this BlogEntry
+	 *
+	 * @return array Associative array of lowercase tag to native case tags
+	 * @deprecated since version 2.0
+	 */
+	public function TagNames() {
+		$tags = preg_split("/\s*,\s*/", trim($this->Tags));
+		$results = array();
+		foreach($tags as $tag) {
+			if($tag) $results[mb_strtolower($tag)] = $tag;
+		}
+		return $results;
+	}
+
+	public function canCreate($member = null) {
+		// Deprecated
+		return false;
+	}
+
+	public function up() {
+		// Migrate tags
+		foreach($this->TagNames() as $tag) {
+			// Skip if tag already assigned
+			if($this->Tags()->filter('Title', $tag)->count()) continue;
+
+			// Create tag
+			$tagObject = new BlogTag();
+			$tagObject->Title = $tag;
+			$tagObject->BlogID = $this->ParentID;
+			$tagObject->write();
+			$this->Tags()->add($tagObject);
+		}
+
+		// Update fields
+		$this->PublishDate = $this->Date;
+		if($this->ClassName === 'BlogEntry') {
+			$this->ClassName = 'BlogPost';
+			$this->write();
+		}
+	}
+
+	public function requireDefaultRecords() {
+		parent::requireDefaultRecords();
+
+		if(BlogMigrationTask::config()->run_during_dev_build) {
+			$task = new BlogMigrationTask();
+			$task->up();
+		}
+	}
+}
+
+/**
+ * @deprecated since version 2.0
+ */
+class BlogEntry_Controller extends BlogPost_Controller {
+}
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,8 @@ env:
     - DB=MYSQL CORE_RELEASE=3.1
 
 matrix:
+  allow_failures:
+    - php: hhvm-nightly
   include:
     - php: 5.6
       env: DB=MYSQL
@@ -29,4 +31,4 @@ before_script:
   - cd ~/builds/ss
 
 script:
-  - phpunit silverstripe-blog/tests/
+  - vendor/bin/phpunit blog/tests/
--- a/code/compat/pages/BlogHolder.php
+++ b/code/compat/pages/BlogHolder.php
@@ -1,0 +1,37 @@
+<?php
+
+/**
+ * @deprecated since version 2.0
+ */
+class BlogHolder extends BlogTree implements MigratableObject {
+
+	private static $hide_ancestor = 'BlogHolder';
+	
+	private static $db = array(
+		'AllowCustomAuthors' => 'Boolean',
+		'ShowFullEntry' => 'Boolean',
+	);
+
+	private static $has_one = array(
+		'Owner' => 'Member',
+	);
+
+	public function canCreate($member = null) {
+		// Deprecated
+		return false;
+	}
+
+	public function up() {
+		if($this->ClassName === 'BlogHolder') {
+			$this->ClassName = 'Blog';
+			$this->write();
+		}
+	}
+}
+
+/**
+ * @deprecated since version 2.0
+ */
+class BlogHolder_Controller extends BlogTree_Controller {
+	
+}
--- a/code/compat/pages/BlogTree.php
+++ b/code/compat/pages/BlogTree.php
@@ -1,0 +1,32 @@
+<?php
+
+/**
+ * @deprecated since version 2.0
+ */
+class BlogTree extends Blog implements MigratableObject {
+
+	private static $hide_ancestor = 'BlogTree';
+
+	private static $db = array(
+		'Name' => 'Varchar(255)',
+		'LandingPageFreshness' => 'Varchar',
+	);
+
+	public function canCreate($member = null) {
+		// Deprecated
+		return false;
+	}
+
+	public function up() {
+		if($this->ClassName === 'BlogTree') {
+			$this->ClassName = 'Blog';
+			$this->write();
+		}
+	}
+}
+
+/**
+ * @deprecated since version 2.0
+ */
+class BlogTree_Controller extends Blog_Controller {
+}
--- a/code/compat/tasks/BlogMigrationTask.php
+++ b/code/compat/tasks/BlogMigrationTask.php
@@ -1,0 +1,73 @@
+<?php
+
+/**
+ * Description of BlogMigrationTask
+ *
+ * @author dmooyman
+ */
+class BlogMigrationTask extends MigrationTask {
+
+	/**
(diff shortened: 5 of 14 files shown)


code/model/Blog.php
```diff
@@ -266,21 +266,21 @@ public function rss() {
...
-
-		// If pagination is set to '0' then no pagination will be shown.
-		if($this->PostsPerPage > 0) $posts->setPageLength($this->PostsPerPage);
-		else $posts->setPageLength($this->getBlogPosts()->count());
-
-		$start = $this->request->getVar($posts->getPaginationGetVar());
-		$posts->setPageStart($start);
-
-		return $posts;
+		// Set page size
```
micmania1: I think regardless of whether or not all results are returned, it should return PaginatedList for simplicity of the API.

templates/Layout/BlogPost.ss
```diff
@@ -0,0 +1,17 @@
+<% require css(themes/simple_blog/css/blog.css) %>
```
micmania1: The path needs updated. I think its fair to assume the folder name here rather than muddying the waters with PHP requirements.

templates/widgets/BlogArchiveWidget.ss
```diff
@@ -1,7 +1,12 @@
 <% if Archive %>
```
micmania1: Should be changed to $Archive

templates/widgets/BlogArchiveWidget.ss
```diff
@@ -1,7 +1,12 @@
 <% if Archive %>
 	<ul>
 		<% loop Archive %>
```
micmania1: and here.

templates/Layout/Blog.ss
```diff
@@ -0,0 +1,64 @@
...
+			<% end_if %>
+			<a class="readMoreLink" href="$Link" title="Read more about &quot;{$Title}&quot;">Read more about &quot;{$Title}&quot;...</a>
+		</li>
+		<% end_loop %>
+	</ul>
+	<% else %>
+		<div class="content"><p>No Posts</p></div>
+	<% end_if %>
+
+	<% if PaginatedList.MoreThanOnePage %>
```
micmania1: Have you tested this without any blog posts. (ie. the division by zero error we seen the other day).

templates/Layout/Blog.ss
```diff
@@ -0,0 +1,64 @@
...
+		<% else_if CurrentTag %>
+			Tag: $CurrentTag.Title
+		<% else_if CurrentCategory %>
+			Category: $CurrentCategory.Title
+		<% else %>
+			$Title
+		<% end_if %>
+	</h1>
+
+	<% if PaginatedList %>
```
micmania1: $PaginatedList

templates/Layout/Blog.ss
```diff
@@ -0,0 +1,64 @@
...
+	<ul id="SearchResults">
+		<% loop PaginatedList %>
```
micmania1: and here.

templates/Layout/Blog.ss
```diff
@@ -0,0 +1,64 @@
+<% require css(themes/simple_blog/css/blog.css) %>
+
+<% include BlogSideBar %>
+<div class="unit size3of4 lastUnit searchResults">
+	<h1>
+		<% if ArchiveYear %>
```
micmania1: $ArchiveYear

templates/Layout/Blog.ss
```diff
@@ -0,0 +1,64 @@
+<% require css(themes/simple_blog/css/blog.css) %>
+
+<% include BlogSideBar %>
+<div class="unit size3of4 lastUnit searchResults">
+	<h1>
+		<% if ArchiveYear %>
+			Archive: <% if ArchiveDay %>$ArchiveDate.Nice<% else_if ArchiveMonth %>$ArchiveDate.format("F, Y")<% else %>$ArchiveDate.format(Y)<% end_if %>
+		<% else_if CurrentTag %>
```
micmania1: $CurrentTag

templates/Layout/Blog.ss
```diff
@@ -0,0 +1,64 @@
+<% require css(themes/simple_blog/css/blog.css) %>
+
+<% include BlogSideBar %>
+<div class="unit size3of4 lastUnit searchResults">
+	<h1>
+		<% if ArchiveYear %>
+			Archive: <% if ArchiveDay %>$ArchiveDate.Nice<% else_if ArchiveMonth %>$ArchiveDate.format("F, Y")<% else %>$ArchiveDate.format(Y)<% end_if %>
+		<% else_if CurrentTag %>
+			Tag: $CurrentTag.Title
+		<% else_if CurrentCategory %>
```
micmania1: etc.

templates/Includes/EntryMeta.ss
```diff
@@ -0,0 +1,17 @@
+<p>
+	This entry was
+	<% if Categories %>
```
micmania1: $Categories

templates/Includes/EntryMeta.ss
```diff
@@ -0,0 +1,17 @@
+<p>
+	This entry was
+	<% if Categories %>
+		posted in
+		<% loop Categories %>
```
micmania1: and here.

templates/Includes/EntryMeta.ss
```diff
@@ -0,0 +1,17 @@
+<p>
+	This entry was
+	<% if Categories %>
+		posted in
+		<% loop Categories %>
+			<a href="$Link" title="$Title">$Title</a><% if not Last %>, <% else_if Up.Tags %>, <% else %> and<% end_if %>
+		<% end_loop %>
+	<% end_if %>
+	<% if Tags %>
```
micmania1: $Tags

templates/Includes/EntryMeta.ss
```diff
@@ -0,0 +1,17 @@
...
+	This entry was
+	<% if Categories %>
+		posted in
+		<% loop Categories %>
+			<a href="$Link" title="$Title">$Title</a><% if not Last %>, <% else_if Up.Tags %>, <% else %> and<% end_if %>
+		<% end_loop %>
+	<% end_if %>
+	<% if Tags %>
+		tagged
+		<% loop Tags %>
```
micmania1: and here.

code/model/Blog.php
tractorcow: I guess

templates/Layout/Blog.ss
tractorcow: Yep, but i'll revert the API anyway. It's always ok to call non-existent methods in templates.

templates/Layout/BlogPost.ss
tractorcow: Ah yeah. We should use themedCSS helper anyway.

code/compat/pages/BlogEntry.php
```diff
@@ -0,0 +1,72 @@
...
+		}
+		return $results;
+	}
+
+	public function canCreate($member = null) {
+		// Deprecated
+		return false;
+	}
+
+	public function up() {
```
micmania1: Hooks should be provided in up() methods so that devs can upgrade bespoke code.

code/compat/pages/BlogEntry.php
```diff
@@ -0,0 +1,72 @@
...
+		"Tags" => "Text"
+	);
+
+	/**
+	 * Safely split and parse all distinct tags assigned to this BlogEntry
+	 *
+	 * @return array Associative array of lowercase tag to native case tags
+	 * @deprecated since version 2.0
+	 */
+	public function TagNames() {
```
micmania1: Can this be made private so that its not used outside of this class?

code/compat/pages/BlogEntry.php
tractorcow: Sure, that sounds doable.

code/compat/pages/BlogEntry.php
```diff
@@ -0,0 +1,72 @@
...
+		if($this->ClassName === 'BlogEntry') {
+			$this->ClassName = 'BlogPost';
+			$this->write();
+		}
+	}
+
+	public function requireDefaultRecords() {
+		parent::requireDefaultRecords();
+
+		if(BlogMigrationTask::config()->run_during_dev_build) {
```
micmania1: I don't think we should give the option to run via dev/build. The more decoupled from the core it is makes it easier to remove later.

micmania1:
I've left some comments - mostly little things.

templates/Layout/BlogPost.ss
micmania1: I don't think themedCSS will work for modules.

code/compat/pages/BlogEntry.php
tractorcow: All of the code is in the same compat folder, so removing it later is as easy as deleting that folder.

code/compat/pages/BlogEntry.php
tractorcow: That would defeat the purpose of deprecating a method, if you're going to immediately make it inaccessible. The idea is to provide a graceful upgrade process.

templates/Layout/BlogPost.ss
tractorcow: That's exactly what it's for. :) The second parameter is the module name.

micmania1: I never knew that :smile: I always just used <% require css %>

code/compat/pages/BlogEntry.php
micmania1: Ok, just a bit confusing. To my module its new code that's being immediately added and deprecated, but to this blog module its old code that's being gracefully deprecated.

code/model/Blog.php
tractorcow: I've made a PR to disable pagination within the list itself. https://github.com/silverstripe/silverstripe-framework/pull/3970

The existing code crashes if paging = 0 and num-items = 0, so I'll just fix that then.

tractorcow:
I've now removed all templates from this, and updated the travis / unit tests.

Template PR i now at https://github.com/silverstripe/silverstripe-blog/pull/148

code/compat/pages/BlogEntry.php
tractorcow: I've added the hook directly into the upgrade task below.

tractorcow:
Travis passes although scrutuniser failed.

PR #145, opened 2015-02-20
Added templates to blog module

Makreig:
This new module doesn't have any default templates so i have added some.

### DIFF
--- a/css/blog.css
+++ b/css/blog.css
@@ -1,0 +1,3 @@
+.no-sidebar .content-container.size3of4 {
+    width: 75%;
+}
--- a/templates/Includes/BlogPostSummary.ss
+++ b/templates/Includes/BlogPostSummary.ss
@@ -1,0 +1,17 @@
+<div class="post-summary">
+	<% if $FeaturedImage %>
+		<p class="post-image"><a href="$Link">
+			$FeaturedImage.setWidth(795)
+		</a></p>
+	<% end_if %>
+	<h2><a href="$Link">$Title</a></h2>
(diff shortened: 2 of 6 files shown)


mlewis-everley:
Is scrutinisher setup correctly for the blog? It seems to be erroring on trying to find the lumberjack module?

dhensby:
Looks like all the composer stuff is messed up

tractorcow:
There is also an existing PR (awaiting fixing some feedback though) at https://github.com/silverstripe/silverstripe-blog/pull/144, but we can probably collaborate to make a good theme from both. :)

tractorcow:
And yep travis-ci tests from the old repo haven't actually been migrated across properly.

mlewis-everley:
Hmm, would it make more sense for the templates to be managed via a seperate pull request to migration tasks?

I am just thinking that if there are issues with the templates, it will slow down merging the migration tasks and vice versa?

tractorcow:
Yeah I can do that.

mlewis-everley:
So, shall we break out the templates from #144, merge with ours and update this pull request?

tractorcow:
Good idea, although I won't have much time to work on this PR this week.

mlewis-everley:
I created a new pull request and merged in this #144. The new request is #146. Not sure if that was the best way to go about this, but it should now have all the code in one place...

tractorcow:
I've added to @mlewis-everley 's PR at https://github.com/silverstripe/silverstripe-blog/pull/148

tractorcow:
Closed ith #148

PR #146, opened 2015-02-27
Break out templates and expand

mlewis-everley:
I have merged the pull request #144, removed the migration tasks and legacy model classes and then merged in the code from @Makreig.

Not sure if this is the best way to do this?


tractorcow:
Wow this is great work. I'm reviewing it now.

It's great to see other contributors working together on code. :)

tractorcow:
I've added some fixes (as documented) and squashed at https://github.com/silverstripe/silverstripe-blog/pull/148 for review.

mlewis-everley:
No probs, I think the blog module is the core module (outside of the cms) that we use the most, and is really in need of an upgrade, so happy to provide any help we can give :-)

tractorcow:
Closed with #148

PR #148, opened 2015-03-05
Break out templates and expand

tractorcow:
I've reviewed and squashed https://github.com/silverstripe/silverstripe-blog/pull/146, and added a few fixes based on @micmania1's feedback on https://github.com/silverstripe/silverstripe-blog/pull/144

- PaginatedList still uses PaginatedList class to keep API consistent
- All templates from @Makreig have been more or less unchanged, with only fixes made to the template syntax (missing $, "", etc).
- Use themedCSS for blog css file (to allow extension by user code / themes)
- Add XML default encoding to $Excerpt (xss vulnerability)

### DIFF
--- a/code/model/Blog.php
+++ b/code/model/Blog.php
@@ -274,8 +274,12 @@ class Blog_Controller extends Page_Controller {
 		$posts = new PaginatedList($this->blogPosts);
 
 		// If pagination is set to '0' then no pagination will be shown.
-		if($this->PostsPerPage > 0) $posts->setPageLength($this->PostsPerPage);
-		else $posts->setPageLength($this->getBlogPosts()->count());
+		if($this->PostsPerPage > 0) {
+			$posts->setPageLength($this->PostsPerPage);
+		} else {
+			$pageSize = $this->getBlogPosts()->count() ?: 99999;
+			$posts->setPageLength($pageSize);
+		}
 
 		$start = $this->request->getVar($posts->getPaginationGetVar());
 		$posts->setPageStart($start);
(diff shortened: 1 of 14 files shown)


code/model/Blog.php
```diff
@@ -274,8 +274,12 @@ public function PaginatedList() {
 		$posts = new PaginatedList($this->blogPosts);
 
 		// If pagination is set to '0' then no pagination will be shown.
-		if($this->PostsPerPage > 0) $posts->setPageLength($this->PostsPerPage);
-		else $posts->setPageLength($this->getBlogPosts()->count());
+		if($this->PostsPerPage > 0) {
+			$posts->setPageLength($this->PostsPerPage);
+		} else {
+			$pageSize = $this->getBlogPosts()->count() ?: 99999;
```
dhensby: why do we need the magic 99999 number?

assertchris: `PHP_INT_MAX` for maximum magic!

dhensby: That's the kind of magic I like! :P

tractorcow: When there are no blog posts setPageLength(0) crashes the site due to divide by zero.

I got annoyed and made https://github.com/silverstripe/silverstripe-framework/pull/3970 trying to fix this.

micmania1: It could just be set to 1 too... There's no posts so it doesn't matter :) Maybe add a comment so future contributors know why its been done and maybe a @todo to remove when your PR is merged.

PR #152, opened 2015-03-05
BUG Fix Type field conflict with core Widget field

tractorcow:
Type is used for the widget classname


micmania1:
It needs a migration script for upgrading as other DO's have.

tractorcow:
Yeah I'll include it... I'm working on a PR for widgets at the moment so I will update this at the same time.

tractorcow:
I've update this as suggested.

I've also updated it to work nicely with widgets module, should https://github.com/silverstripe/silverstripe-widgets/pull/84 be merged.

tractorcow:
Thanks. :)

PR #153, opened 2015-03-05
Add extension hooks to blog entry form

dhensby:
re-run of #126


tractorcow:
We need to fix travis from master -> 3

PR #155, opened 2015-03-10
Update nl.yml

derkgort:
The dutch "Aan" is not the right wordt to use, "op" is.


assertchris:
Thanks for this but it would be better to add translation stuff on [Transifex](https://www.transifex.com/projects/p/silverstripe-blog). This file will just get overridden the next time we update from Transifex...

PR #156, opened 2015-03-10
Dutch language 2

derkgort:
This adds a few texts to the dutch language yml.

### DIFF
--- a/lang/nl.yml
+++ b/lang/nl.yml
@@ -17,7 +17,7 @@ nl:
     COMMENTS: "Reacties"
     EDITTHIS: "Bewerk deze post"
     POSTEDBY: "Auteur"
-    POSTEDON: "op"
+    POSTEDON: "Op"
     TAGS: "Tags:"
     UNPUBLISHTHIS: "onpubliceer deze post"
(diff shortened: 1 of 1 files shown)


lang/nl.yml
```diff
@@ -17,7 +17,7 @@ nl:
     COMMENTS: "Reacties"
     EDITTHIS: "Bewerk deze post"
     POSTEDBY: "Auteur"
-    POSTEDON: "Aan"
+    POSTEDON: "op"
```
dhensby: I think we need to keep the capitalisation ie: it should be "Op"

derkgort: That's OK. Thanks

derkgort: Do you need a new request?

dhensby: You can just amend the commits in this branch and the PR will update

dhensby: Note this needs amending in both places you've changed it

derkgort: Done

tractorcow:
Instead of merging it in here, perhaps we should consolidate with https://www.transifex.com/projects/p/silverstripe-blog/viewstrings/#nl/master?

PR #157, opened 2015-03-10
Fixed blogentry migration

heffaklump90:
Forced a regeneration of class manifest. The call to BlogEntry::get() returned an empty set without it
Also added an explicit publish of the already published blog entries, otherwise I had to go through and publish them all again.

### DIFF
--- a/code/compat/pages/BlogEntry.php
+++ b/code/compat/pages/BlogEntry.php
@@ -52,6 +52,9 @@ class BlogEntry extends BlogPost implements MigratableObject {
 		if($this->ClassName === 'BlogEntry') {
 			$this->ClassName = 'BlogPost';
 			$this->write();
+			if($this->isPublished()){
+				$this->doPublish();
+			}
 		}
 	}
 
--- a/code/compat/tasks/BlogMigrationTask.php
+++ b/code/compat/tasks/BlogMigrationTask.php
@@ -24,6 +24,8 @@ class BlogMigrationTask extends MigrationTask {
 	}
 
 	public function up() {
+		SS_ClassLoader::instance()->getManifest()->regenerate(false);
+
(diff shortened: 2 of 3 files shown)


tractorcow:
Hi @heffaklump90 thanks for this work. I think there might be a few problems still with the solution I'd like to iron out:

 * There can be cases where there is content on draft and different content on live. In your fix, the live content would always be overwritten with the draft version, which might not be the desired outcome.
 * There's a duplication between calling SiteTree->doPublish and writing records in the live stage (see https://github.com/silverstripe/silverstripe-blog/blob/a81ff7067d1ad20d31b3259f1083e264ca1659da/code/compat/tasks/BlogMigrationTask.php#L34). Why does the above code not publish the live entry where doPublish does? (I understand there are some nuance differences between the two, so it's worth me testing this again. Perhaps the Live stage should be migrated first?
* I think if updating the module to the new version it's probably worth using flush=all over trying to force regeneration of the class cache explicitly. This could be a performance / stability issue if this task is called automatically every dev/build (when sometimes flushing the class manifest isn't always desired). Is your fix an attempt to negate the need to add the flush parameter?

heffaklump90:
* Hi @tractorcow, you are correct in the first point. 
* For your second point I have not looked into it deeply, my main problem was the BlogEntry records werent fetched from the db, even when flushing. Since it's a migration task, I will probably just do the publish thing in an extension in onAfterUp instead. I don't have the problem with unpublished changes.
* I didn't get the migration to work with the flush parameter either - or at least I thought so. I have tried it afterwards and it works, so my fix is probably just waste :-)

tractorcow:
I'm going to consider point 2 an open issue still... I will investigate to make sure you haven't encountered a legitimate bug. :)

PR #158, opened 2015-03-10
[WIP] Adding authors (credit) to posts

assertchris:


code/model/BlogPost.php
```diff
@@ -123,6 +144,18 @@ public function getCMSFields() {
 	**/
 	protected function onBeforeWrite() {
 		parent::onBeforeWrite();
+
+		if(isset($_POST["AuthorFirstName"]) or isset($_POST["AuthorLastName"])) {
```
dhensby: Don't manipulate the superglobals.

These will likely be loaded onto the object and available as `$this->AuthorFirstName` or you can get them off the current request on the controller.

code/model/BlogPost.php
```diff
@@ -123,6 +144,18 @@ public function getCMSFields() {
 	**/
 	protected function onBeforeWrite() {
 		parent::onBeforeWrite();
+
+		if(isset($_POST["AuthorFirstName"]) or isset($_POST["AuthorLastName"])) {
+			$author = new Member();
```
dhensby: Hmm, creating a member so simply is likely to cause issues for anyone with custom requirements / validation on their `Member` object.

Does an Author have to be a `Member`? Perhaps an `Author` object is more appropriate that may or maynot be linked to a `Member` object itself?

assertchris: Good point! What do you think, @tractorcow?

dhensby:
This needs test(s), I think

assertchris:
I agree. Should I mimic the Behat setup from another module? Currently there are only PHPUnit tests, which I think would be of little value for this feature.

dhensby:
Well, I think you need unit tests to ensure that when the member info is submited the relations are set up as expected - that there aren't collisions might be another scenario to cover?

Behat tests would probably be a bit out of scope for this PR if there are none in the module yet?

assertchris:
Cool - I'll add those unit tests and pick up the Behat stuff in another PR. 

I've renamed this PR to reflect the working state.

code/model/BlogPost.php
```diff
@@ -100,7 +101,27 @@ public function getCMSFields() {
...
 					$self->Parent()->Tags()->map()->toArray()
-				)->setMultiple(true)
+				)->setMultiple(true),
+				ListboxField::create(
+					"Authors",
+					_t("BlogPost.Authors", "Authors"),
+					Member::get()->map()->toArray()
+				)->setMultiple(true),
+				FieldGroup::create(
+					LabelField::create(null, 'Add New Author')->addExtraClass("new-author-label left"),
```
tractorcow: For authors which don't exist I think it's enough to just store the name as a plain textfield.

assertchris: So along the lines of what @dhensby suggested - a new object potentially linked to a Member.

tractorcow: I'm not sure it needs to be an object, especially if the user doesn't have a member record. It's only a label really.

tractorcow:
I think this feature has already been implemented and merged, so can I close this PR?

PR #162, opened 2015-03-23
Add member profiles to blog

flashbackzoo:


code/extensions/BlogMemberExtension.php
```diff
@@ -0,0 +1,69 @@
...
+ * This class is responsible for add Blog specific behaviour to Members.
+ *
+ * @package silverstripe
+ * @subpackage blog
+ *
+ **/
+class BlogMemberExtension extends DataExtension {
+
+    private static $db = array(
+        'Slug' => 'Varchar',
```
tractorcow: Please use tabs for indentation not spaces. :)

code/extensions/BlogMemberExtension.php
```diff
@@ -0,0 +1,69 @@
...
+        'Summary' => 'Text'
+    );
+
+    private static $has_one = array(
+        'Image' => 'Image'
```
tractorcow: This is a bit generic. What is the image for? Think of an application where a member might have many images. Should it be BlogProfileImage?

code/extensions/BlogMemberExtension.php
```diff
@@ -0,0 +1,69 @@
...
+    /**
+     * Returns TRUE if this object has a Slug value that does not conflict with any other objects.
+     * 
+     * @return bool
+     */
+    public function validSlug() { 
+        $isValid = true;
+        $members = Member::get();
+
+        foreach ($members as $member) {
```
tractorcow: Instead of looping every member every check, simply use the filter to find any matching member.

```php
$conflict = Member::get()->filter('Slug', $this->owner->Slug)->exclude('ID', $this->ID)->count();
return !$conflict;
```

code/model/Blog.php
```diff
@@ -473,7 +475,58 @@ public function index() {
...
+		return null;
+	}
+
+	/**
+	 * Get posts related to the current Member profile
+	 * 
+	 * @return ArrayList|null
+	**/
+	public function getCurrentProfilePosts() {
+		if($this->getCurrentProfile()) {
```
tractorcow: Or just assign this to a variable and save calling the method twice. :D

code/model/Blog.php
```diff
@@ -473,7 +475,58 @@ public function index() {
...
+			return BlogPost::get()->filterByCallback(function($blogPost) {
+				$profile = $this->getCurrentProfile();
```
tractorcow: You can't use `$this` inside a closure in php 5.3

code/extensions/BlogMemberExtension.php
```diff
@@ -0,0 +1,69 @@
...
+
+    private static $db = array(
+        'Slug' => 'Varchar',
+        'Summary' => 'Text'
+    );
+
+    private static $has_one = array(
+        'Image' => 'Image'
+    );
+
```
tractorcow: If you add a `belongs_many_many = array('AuthoredPosts' => 'BlogPost')` then you can use this relation directly in your blog filter below.

`$posts = $member->AuthoredPosts();`

code/model/Blog.php
```diff
@@ -473,7 +475,58 @@ public function index() {
...
+
+				if ($blogPost->ParentID == $this->ID && $blogPost->Authors()->filter('ID', $profile->ID)->First()) {
```
tractorcow: This whole condition really needs to be inside a single sql query. If you add the relation to member I suggested above, you can simplify this code to a single query:

```
return $profile->AuthoredPosts()->filter('ParentID', $this->owner->ID);
```

templates/Includes/EntryMeta.ss
```diff
@@ -26,7 +26,12 @@
 	<% if $Authors || $AuthorNames %>
 		<%t Blog.By "by" %>
 		<% if $Authors %><% loop $Authors %>
-			$Name.XML<% if not $Last || $Up.AuthorNames %>,<% end_if %>
+			<% if $Slug %>
```
tractorcow: Can we call this field `$URLSegment` so that it's consistent with `SiteTree.URLSegment`?

templates/Includes/EntryMeta.ss
```diff
@@ -26,7 +26,12 @@
 	<% if $Authors || $AuthorNames %>
 		<%t Blog.By "by" %>
 		<% if $Authors %><% loop $Authors %>
-			$Name.XML<% if not $Last || $Up.AuthorNames %>,<% end_if %>
+			<% if $Slug %>
+                <a href="{$Up.Parent.Link}profile/$Slug">$Name.XML</a>
```
tractorcow: We could make a method in PHP that takes a single parameter (the author slug/urlsegment) and does the string concatenation in a URL safe way. Doing it in the template can be risky.

```
public function ProfileLink($urlsegment) {
    return Controller::join_links($this->Link(), 'profile', $urlsegment);
}
```

That way you get all the extra XSS protection and URL rewriting checks from join_links method.

code/extensions/BlogMemberExtension.php
```diff
@@ -0,0 +1,69 @@
...
+ *
+ * @package silverstripe
+ * @subpackage blog
+ *
+ **/
+class BlogMemberExtension extends DataExtension {
+
+    private static $db = array(
+        'Slug' => 'Varchar',
+        'Summary' => 'Text'
```
tractorcow: Have you checked how the CMS Fields for this are generated? It might be worth cleaning it up in updateCMSFields.

templates/Layout/Blog_Profile.ss
```diff
@@ -0,0 +1,23 @@
+<% require themedCSS('blog', 'blog') %>
```
tractorcow: The naming of the file should be Blog_profile.ss (since the action is lowercase `profile`).

code/extensions/BlogMemberExtension.php
```diff
@@ -0,0 +1,70 @@
...
+ * This class is responsible for add Blog specific behaviour to Members.
+ *
+ * @package silverstripe
+ * @subpackage blog
+ *
+ **/
+class BlogMemberExtension extends DataExtension {
+
+	private static $db = array(
+		'URLSegment' => 'Varchar',
```
tractorcow: This URLSegment field is showing up in the CMS. The user shouldn't be able to edit this themselves.

code/model/Blog.php
```diff
@@ -473,7 +475,64 @@ public function index() {
...
+		return null;
+	}
+
+	/**
+	 * Get a link to a Member profile.
+	 * 
+	 * @param urlSegment
+	 * @return String
+	 */
+	public function getProfileLink($urlSegment) {
```
tractorcow: Needs to be named `ProfileLink` not `getProfileLink` since it's a method that requires a parameter and not a property getter.

Also needs to be on the `Blog` dataobject not the `Blog_Controller` since $blogPost->Parent() only returns the model instance.

code/extensions/BlogMemberExtension.php
dhensby: And it should have a length defined

code/extensions/BlogMemberExtension.php
```diff
@@ -0,0 +1,76 @@
...
+	}
+
+	/**
+	 * Generate a unique URL segment based on the Member's name.
+	 * 
+	 * @return string Generated URL segment.
+	 */
+	public function generateURLSegment() {
+		$filter = URLSegmentFilter::create();
+		$name = $this->owner->FirstName . ' ' . $this->owner->Surname;
```
dhensby: `$this->owner->getName()`?

code/extensions/BlogMemberExtension.php
```diff
@@ -0,0 +1,76 @@
...
+	public function validURLSegment() { 
+		$conflict = Member::get()->filter('URLSegment', $this->owner->URLSegment);
+
+		// If the Member we're checking against is saved, exclude them from the check.
+		// i.e. don't conflict against yourself.
+		if ($this->owner->ID) {
+			$conflict = $conflict->exclude('ID', $this->owner->ID);
+		}
+
+		return $conflict->count() == 0;
```
dhensby: `return $conflict->exists()`

code/extensions/BlogMemberExtension.php
```diff
@@ -0,0 +1,76 @@
...
+
+	private static $belongs_many_many = array(
+		'AuthoredPosts' => 'BlogPost'
+	);
+
+	public function onBeforeWrite() {
+		// Generate a unique URL segment for the Member.
+		$count = 1;
+
+		$this->owner->URLSegment = $this->generateURLSegment();
```
dhensby: Should this really be regenerated on every write? The URLs don't have any history (do they?) and so changing URLs will lead to 404s and dead links.

Once the URL is set (if there's no history) it shouldn't really change

dhensby:
probably want a (unique) index of the MemberExtension for the `URLSegment` too?

tractorcow:
Too late it's merged. :)

code/extensions/BlogMemberExtension.php
tractorcow: That's true, but likely it won't ever change unless two people have the same name.

flashbackzoo:
https://github.com/silverstripe/silverstripe-blog/pull/164

code/extensions/BlogMemberExtension.php
dhensby: Or if someone changes their name...

PR #164, opened 2015-03-24
Minor improvements to BlogMemberExtension

flashbackzoo:


code/extensions/BlogMemberExtension.php
```diff
@@ -10,7 +10,7 @@
 class BlogMemberExtension extends DataExtension {
 
 	private static $db = array(
-		'URLSegment' => 'Varchar',
+		'URLSegment' => 'Varchar(50)',
```
tractorcow: If a name is more than 50 characters, you will en up with a database error.

You'll need to do a substr to limit the max characters in your generateURLSegment :)

dhensby:
Unique index on `URLSegment`?

Only generate a `URLSegment` if there isn't one already?

assertchris:
@flashbackzoo want to pursue this, or should we close?

flashbackzoo:
close

PR #165, opened 2015-03-25
BUG Fix editors not being able to create categories / tags

tractorcow:


tractorcow:
Should have updated the tests :)

dhensby:
So this is removing a feature where editors can't edit specific categories? Are there any docs that need updating?

tractorcow:
No it's fixing a bug in the logic. You can't call canCreate on an instance that's already been created. In canCreate you can't see who the editors are, because it's normally called on singletons that aren't attached to any blog.

The real "fix" would be to introduce the context in canCreate like we did for SiteTree, but that would necessitate additional changes to core.

PR #167, opened 2015-03-30
Remove 'Blog' from tags/categories buttons

scott1702:
Remove the word 'Blog' from the add tag/category buttons. This word is unnecessary as the user is already aware of what they are doing.


tractorcow:
Sure, sounds reasonable.

How about "Blog Categories" => "Categories" as well? :P

PR #168, opened 2015-03-31
BUG Fix incorrect reference to extension member ID

tractorcow:
$this->ID doesn't work in an extension.


dhensby:
An easy mistake to make :)

PR #169, opened 2015-04-01
Fixed author credits in blog

assertchris:


code/model/BlogPost.php
```diff
@@ -3,15 +3,15 @@
 /**
  * An indivisual blog post.
  *
- * @package silverstripe
+ * @package    silverstripe
```
tractorcow: What formatting convention is this? Just curious.

tractorcow:
The problem with this is that it feels like we are doing front-end formatting logic in the model layer, when it should be left in the view level. It makes it very hard for templates to override this behaviour. We should try to be as MVC as possible. :)

Perhaps the getStaticAuthors would be useful to convert the textfield into a list that can be iterated in the template, however. To do this you'd just return an ArrayList of all the split names. :)

There's also a lack of translation ability here. Not all locales will use "and" as their join. This is why it's better to leave it to the templates to determine.

Anyway, good progress so far. :)

code/model/BlogPost.php
```diff
@@ -319,41 +319,137 @@ public function getMonthlyArchiveLink($type = "day") {
...
+	 *
+	 * @param array $items
+	 *
+	 * @return array
+	 */
+	protected function itemsWithJoins(array $items) {
+		$count = count($items);
+
+		for($i = 0; $i < $count; $i++) {
+			if($count === 2 && $i === 1) {
```
tractorcow: This logic was already in the template, but putting it here means it's now less easily overridden. All you need to add is a `<% if $Last && not $First %>and<% end_if %>` before the author (oxford comma pending discussion).

Have a look at http://doc.silverstripe.org/en/developer_guides/templates/syntax/#position-indicators for reference on how the various iterator variables can be used.

code/model/BlogPost.php
```diff
@@ -319,41 +319,137 @@ public function getMonthlyArchiveLink($type = "day") {
...
 
+	/**
+	 * Resolves dynamic and static authors with appropriate join data.
+	 *
+	 * @return ArrayList
+	 */
+	public function getCredits() {
+		$dynamic = $this->getDynamicAuthors();
+
+		$static = $this->getStaticAuthors();
```
tractorcow: How about simplifying (and maybe a nice sort on name?)

```php
return ArrayList::create()
->merge($this->Authors())
->merge($this->getStaticAuthors())
->sort('Name');
```

code/model/BlogPost.php
```diff
@@ -319,41 +319,137 @@ public function getMonthlyArchiveLink($type = "day") {
...
+			)
+		);
+	}
+
+	/**
+	 * Resolves all dynamic authors linked to this post.
+	 *
+	 * @return array
+	 */
+	protected function getDynamicAuthors() {
```
tractorcow: This method is possibly not needed. :)

code/model/BlogPost.php
```diff
@@ -319,41 +319,137 @@ public function getMonthlyArchiveLink($type = "day") {
...
+		return $items;
+	}
+
+	/**
+	 * Resolves all static authors linked to this post.
+	 *
+	 * @return array
+	 */
+	protected function getStaticAuthors() {
+		$items = [];
```
tractorcow: I could simplify to
```php
return ArrayList::create(preg_split('/\s*,\s*/', $this->AuthorNames));
```

In a template, it's perfectly fine to refer to variables on an object that don't exist, so you can safely ignore values which would otherwise evaluate to false.

$URL, $URLSegment, etc will all be empty and safe to use even for objects without those defined.

code/model/BlogPost.php
```diff
@@ -319,41 +319,137 @@ public function getMonthlyArchiveLink($type = "day") {
...
+			$items[] = $item;
+		}
+
+		return $items;
+	}
+
+	/**
+	 * Resolves all static authors linked to this post.
+	 *
+	 * @return array
```
tractorcow: If it returns an ArrayList instead then that lets users refer to this property in their templates directly.

tractorcow: If it were public that is. :)

code/model/BlogPost.php
assertchris: Merge doesn't return the list, so these can't be chained.

code/model/BlogPost.php
assertchris: How does that help for "not the first and not the last if total items > 2"?

assertchris: ...or, for that matter, !$First and !$Last...

code/model/BlogPost.php
tractorcow: Well you can use a variable then. :) I just wrote it off the top of my head.

code/model/BlogPost.php
tractorcow: You can nest conditions

```
<% if not $First >
  <% if $Count <= '2' or $Last %>
    
  <% end_if %>
<% end_if %>

tractorcow: `<% if not $First or not $Last %>`

tractorcow: Also you can use <% unless %> to flip conditions around if that makes things simpler.

PR #170, opened 2015-04-01
Fixed credits

assertchris:


code/model/BlogPost.php
```diff
@@ -330,7 +330,66 @@ public function getYearlyArchiveLink() {
...
+	 * Resolves dynamic authors linked to this post.
+	 *
+	 * @return ArrayList
+	 */
+	protected function getDynamicCredits()
+	{
+		$items = new ArrayList();
+
+		foreach($this->Authors() as $author) {
+			$item = new ArrayData(array(
```
tractorcow: Ok, one more possible enhancement;

```php
$item = $author->customise(array(
    'URL' => $this->Parent->ProfileLink($author->URLSegment);
));
```

This means that the frontend guys can use $author.GravitarURL, $author.MyCustomDatabaseField, etc in their templates.

customise does similarly to what wrapping an ArrayData does, but allows fallback property actions to access the internal object.

PR #172, opened 2015-04-02
Make BlogPost::isAuthor public

flashbackzoo:
You can't use isAuthor in extensions because it's currently protected.


assertchris:
Why?

tractorcow:
He wants to check if a comment author is also the author of the blog post.

assertchris:
Seems legit. :)

PR #174, opened 2015-04-07
Added tag field

assertchris:


chillu:
Veto'ing this change until we figure out what the direction of the existing `silverstripe/tagfield` is, how this module is different from https://github.com/sheadawson/silverstripe-select2, and how we deal with having two quite similar modules in the `silverstripe` vendor namespace. Off to a meeting now, just making sure this isn't merged beforehand :)

micmania1:
I was actually looking to implement select2 into core in the past to replace chosen, but I hit all sorts of issues (only with chosen - not select2). See here: https://github.com/silverstripe/silverstripe-framework/pull/2891

I think the version included in that module needs updated but +1 for select2.

chillu:
@micmania1 What's the driver for using select2 over chosen? The inline add of new tags seems to be the selling point for select2, right? There's been discussions since 2012 on chosen to get this into the library, and it looks like most contributors there have given up: https://github.com/harvesthq/chosen/pull/166

Looking at it from an open source project longevity perspective, both have a large amount of open bugs for the scope of this UI component (270 for select2 vs. 241 for chosen). It looks like select2 has more traction (3x the contributors), but that's what I thought two years ago about chosen as well when we decided to use the library for core. 

We should strive to remove the ugly hacks in the framework chosen.js "fork" (https://github.com/silverstripe/silverstripe-framework/commit/a73c8bcbf01f10e1f6a9d8a2c9e6a927f978f51f), but that's hardly enough of a reason to replace the whole library approach.

So, I'm open for a core change, but it would need to be well considered beyond "the shiny new thing" :) Is there anything apart from the "add new" feature?

micmania1:
Select2 allowed you to extend and customise it much more easily and solved the issues for which we had to hack away at chosen.

At the time I was doing this, the tag field didn't exist as it does now - it looks to have improved a lot. It used to essentially be a ListboxField but with more extension points (where I needed them).

PR #176, opened 2015-04-09
API Enable comment specific enhancements

tractorcow:
If an author creates a comment include a highlight class function that can be used in templates.

### DIFF
--- a/_config/config.yml
+++ b/_config/config.yml
@@ -1,3 +1,6 @@
+---
+Name: blogconfig
+---
 Member:
   extensions:
     - BlogMemberExtension
--- a/code/extensions/BlogCommentExtension.php
+++ b/code/extensions/BlogCommentExtension.php
@@ -1,0 +1,23 @@
+<?php
+
+/**
+ * Adds Blog specific behaviour to Comment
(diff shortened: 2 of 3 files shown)


_config/config.yml
```diff
@@ -1,3 +1,6 @@
+---
+Name: blogconfig
+---
 Member:
   extensions:
     - BlogMemberExtension
```
assertchris: plz trailing newline. kthxbai

code/extensions/BlogCommentExtension.php
```diff
@@ -0,0 +1,23 @@
...
+
+	/**
+	 * Extra CSS classes for styling different comment types.
+	 * @return string
+	 */
+	public function getExtraClass() {
+		$blogPost = $this->owner->getParent();
+		
+		// Make sure we're dealing with a BlogPost.
+		if (  ($blogPost instanceof BlogPost)
```
assertchris: `is_a` might be better than `instanceof` in the case of subclasses.

_config/config.yml
tractorcow: Fixed.

PR #177, opened 2015-04-09
API Add role-helpers for templates

tractorcow:
This can also be used by other parts of the code, e.g. email notifications, to customise elements for individuals.

### DIFF
--- a/code/model/BlogPost.php
+++ b/code/model/BlogPost.php
@@ -83,6 +83,27 @@ class BlogPost extends Page {
 		return $list->byID($member->ID) !== null;
 	}
 
+	/**
+	 * Determine the role of the given member
+	 * Call be called via template to determine the current user
+	 *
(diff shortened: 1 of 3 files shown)


code/model/BlogPost.php
```diff
@@ -83,6 +83,27 @@ public function isAuthor($member) {
...
 	}
 
+	/**
+	 * Determine the role of the given member
+	 * Call be called via template to determine the current user
+	 *
+	 * E.g. `Hello $RoleOf($CurrentMember.ID)`
+	 *
+	 * @param Member|integer $member
+	 * @return string Author, Editor, Writer, Contributor, or null if no role
```
assertchris: `@return null|string ...`

tractorcow: Ok fixed.

PR #179, opened 2015-04-09
API Add filter description to blog holder

tractorcow:
Supported by the default theme via $MetaTitle

### DIFF
--- a/code/model/Blog.php
+++ b/code/model/Blog.php
@@ -527,6 +527,11 @@ class Blog_Controller extends Page_Controller {
 		'profile/$URLSegment!' => 'profile'
 	);
 
+	private static $casting = array(
+		'MetaTitle' => 'Text',
+		'FilterDescription' => 'Text'
+	);
+
 
 	/** 
 	 * The current Blog Post DataList query.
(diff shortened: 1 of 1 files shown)


dhensby:
Shouldn't lots of this text use `_t` so it can be translated / customised?

code/model/Blog.php
```diff
@@ -527,6 +527,11 @@ class Blog_Controller extends Page_Controller {
 		'profile/$URLSegment!' => 'profile'
 	);
 
+	private static $casting = array(
+		'MetaTitle' => 'Text',
+		'FilterDescription' => 'Text'
+	);
+
```
chillu: Isn't `Text` the default cast in 3.1 anyway?

tractorcow: Default cast isn't respected by default, only if additional actions are called on the resulting object.

E.g.

```$MyField``` will not be cast to Text.

```$MyField.Action``` will be cast to Text, and have Text::Action invoked on it.

tractorcow:
Description has been localised, thanks @dhensby

PR #180, opened 2015-04-12
Update descriptions of user levels

scott1702:
Updated the descriptions of each user level to better represent the permissions of each user.


assertchris:
Good job!

PR #185, opened 2015-04-14
Adjust size of custom summary and remove extra title

tractorcow:
Depends on https://github.com/silverstripe/silverstripe-framework/pull/4072 (a little bit)


tractorcow:
Cleaned up version looks like the below:

image

PR #195, opened 2015-04-22
BUG Backport of https://github.com/silverstripe/silverstripe-framework/pull/4077

tractorcow:
Better support for versions of framework 3.1.12 and older

### DIFF
--- a/css/cms.css
+++ b/css/cms.css
@@ -128,3 +128,8 @@ label.right.toggle-description-correct-right {
 .cms table.ss-gridfield-table tr td.MergeAction select {
   width: 75px;
 }
+
+.cms-content-actions,
+.cms-preview-controls {
+  z-index: 999;
+}
--- a/scss/cms.scss
+++ b/scss/cms.scss
@@ -159,3 +159,8 @@ label.right.toggle-description-correct-right {
 .cms table.ss-gridfield-table tr td.MergeAction select {
 	width: 75px;
 }
+
+.cms-content-actions,
+.cms-preview-controls {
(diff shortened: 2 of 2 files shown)


sminnee:
Good call—this makes things more flexible and work without hassles.

I wonder if it's worth commenting that this won't be necessary if the site is using silverstripe/framework 3.1.13 or higher, so that we know we can delete this code in 18 months time when the blog module it 3.2+ only anyway? ;-)

assertchris:
There are other areas of blog that just won't work without 3.1.13/3.1.x-dev. Merging tags/categories for instance...

tractorcow:
It's fine if it gracefully degrades. In this case, the feature is unusable not simply degraded.

assertchris:
Merging tags and categories doesn't degrade. It doesn't work without JS.

tractorcow:
Oh, is there a way we could disable that option on other framework versions?

Also this discussion has nothing to do with this PR.

assertchris:
It's slightly related. What I'm trying to say is that it doesn't make sense to backport features to make blog 2.0 run on 3.1.12 if blog 2.0 requires 3.1.13 in other places.

tractorcow:
Which PR in framework is necessary for those features? was it https://github.com/silverstripe/silverstripe-framework/pull/4079?

If you install blog module on an older version of framework what functions do you get? I would say a better solution is to review each of these features and make sure they degrade well.

assertchris:
We need to render a custom row for each record in `GridField`. That's how we get the origin parent from which we move children.

PR #197, opened 2015-04-23
Add composer replace

tractorcow:
cc @micmania1 for approval :D Since we are releasing 2.0 soon it pays to prevent any confusion between this and the original fork.


tractorcow:
See https://github.com/silverstripe/silverstripe-lumberjack/pull/5

PR #205, opened 2015-04-28
Update TagField dependency version

assertchris:


composer.json
```diff
@@ -10,7 +10,7 @@
 	"require": {
 		"silverstripe/cms": ">=3.1.0",
 		"silverstripe/lumberjack": "~1.1",
-		"silverstripe/tagfield": "dev-develop"
+		"silverstripe/tagfield": "2.0.x-dev"
```
tractorcow: Would `~2.0.0` work?

assertchris: I don't know, but we'll have to update this in the next 2 weeks anyway.

Issue #221, opened 2015-05-05
Make "Blog Posts" the Default Tab


purplespider:
Moving Blog posts into a GridField makes perfect sense, however moving them into a secondary **Blog Posts** tab is just a step too far for my liking. I know I will have clients that struggle to work out how to find the posts or add a new one, as I've used this convention many times in the past.

When a client clicks on the Blog page in the CMS, they expect to immediately see the blog posts and the *Add new* button, and not the **Main Content** that they very rarely need to change.

When I used this convention for a News section, I had several clients just type out a new News article into the Main Content tab!

Even when you know where to go to find the blog posts, hiding them away on a 2nd tab is adding an extra step.

Can we please look at making the **Blog Posts** tab the first tab to solve these issues? With the infrequently used **Main Content** tab coming after it?

assertchris:
Thanks for pointing this out. We've added it to our internal issue tracker and will assess whether it should be part of the stable blog 2.0 release or come after.

camfindlay:
+1 for this quick win UX improvement. There is a mention on how to move tabs about over at http://stackoverflow.com/questions/5033028/silverstripe-how-do-i-insert-a-tab-before-another-tab if anyone from the community wants to have a crack and submit a PR for the guys to peer review?

PR #223, opened 2015-05-09
Reorder tabs

assertchris:
https://github.com/silverstripe/silverstripe-blog/issues/221


camfindlay:
Looking good. Tested on fresh 3.1.x core. Much nicer workflow with this.

Issue #233, opened 2015-05-13
Adding Widgets to blog page Fatal error


adrian-stein:
Since one of the last commits when adding a widget to a page in the CMS I get the below error:

Fatal error: Call to a member function merge() on a non-object in G:\Nimbler\es\blog\code\widgets\BlogArchiveWidget.php on line 64

if I comment out that line of code I then get this error:

Fatal error: Call to a member function push() on a non-object in G:\Nimbler\es\blog\code\widgets\BlogTagsWidget.php on line 43

If both lines are commented out then the page loads. Any ideas?

assertchris:
Will check it out!

tractorcow:
I think the issue is you are using widgets 1.0, when 1.1 is the one we were testing against.

A fix on the 1.0 branch is available at https://github.com/silverstripe/silverstripe-widgets/pull/89, although in your case I would still recommend you upgrade. ;)

adrian-stein:
Ah. Yeah changed from using dev/master due to the restriction of having to have at least Silverstripe 3.2. I must have gone back to to 1.0 and not 1.1. I have not put it to 1.1 and it works a treat again. Thanks.

Issue #232, opened 2015-05-13
BlogArchiveWidget GroupBy Issue


adrian-stein:
I am getting an error with the widget since the "Clean up" commit (https://github.com/silverstripe/silverstripe-blog/commit/787c02a19d2491ace59948630795e5bc6da40bf0)

It appears changing the quote tags around the groupBy line in the method getArchive() in BlogArchiveWidget.php now breaks the select statement as seen here: http://postimg.org/image/hllern6yn/ 

if I change put them back to the way it was before the commit it appears to work. Are you able to fix it up?

assertchris:
Well spotted - I'll fix this shortly!

tractorcow:
Fixed with #234

PR #234, opened 2015-05-13
Fixed regressions

assertchris:
@adrian-stein: does this fix the errors you are encountering in https://github.com/silverstripe/silverstripe-blog/issues/233 and https://github.com/silverstripe/silverstripe-blog/issues/232?

### DIFF
--- a/code/widgets/BlogTagsWidget.php
+++ b/code/widgets/BlogTagsWidget.php
@@ -40,7 +40,16 @@ class BlogTagsWidget extends Widget {
 	 */
 	public function getCMSFields() {
 		$this->beforeUpdateCMSFields(function ($fields) {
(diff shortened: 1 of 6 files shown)


code/widgets/BlogTagsWidget.php
```diff
@@ -40,7 +40,16 @@ class BlogTagsWidget extends Widget {
 	 */
 	public function getCMSFields() {
 		$this->beforeUpdateCMSFields(function ($fields) {
-			$fields->push(DropdownField::create('BlogID', _t('BlogTagsWidget.Blog', 'Blog'), Blog::get()->map()));
+			if (!$fields) {
```
tractorcow: I think we should drop this. If $fields are null there's a bug elsewhere right? I'm actually happy it crashed earlier because it helped us find a bug in widgets 1.0.

adrian-stein:
Yup!! Thanks guys that got it! :)

PR #235, opened 2015-05-13
Ansi sql encode quotes

tractorcow:
Some feedback from the original cleanup PR I should have complained about but didn't. :D


tractorcow:
Superseded :D

PR #238, opened 2015-05-14
Fix excerpt parsing

flashbackzoo:
Before:
blog-excerpt-before

After:
blog-excerpt-after

### DIFF
--- a/code/model/BlogPost.php
+++ b/code/model/BlogPost.php
@@ -496,7 +496,7 @@ class BlogPost extends Page {
 		 */
 		$content = $this->dbObject('Content');
 
-		return $content->LimitWordCount($wordsToDisplay);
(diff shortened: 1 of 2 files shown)


flashbackzoo:
Fixes https://github.com/silverstripe/silverstripe-blog/pull/238

tractorcow:
good job!

assertchris:
It has a desirable effect, but doesn't fix the underlying problem. Keen to merge this and still deal with the other value getters...

tractorcow:
The core issue is at https://github.com/silverstripe/silverstripe-framework/issues/1381

I think it should probably be fixed in a point release, since it should probably involve some level of api cleanup.

tractorcow:
Really there's no point having both Summary and LimitWordCount that attempt to do the same thing.

PR #253, opened 2015-05-24
Updated Tag Field version

assertchris:
Version change looks strange, but it's because `2.0.x-dev` was a major version off. Only `0.x` releases existed, previously. Just tagged `1.0.0`.


tractorcow:
New version of tagfield is 2.0 not 1.0

PR #256, opened 2015-05-25
Fix migration task

flashbackzoo:
- Published dates were not getting migrated correctly.


code/compat/pages/BlogEntry.php
```diff
@@ -50,10 +50,9 @@ public function up() {
 
 		$this->PublishDate = $this->Date;
 
-		if($this->ClassName === 'BlogEntry') {
-			$this->ClassName = 'BlogPost';
-		}
-		
+		$this->ClassName = 'BlogPost';
```
tractorcow: This should not have been merged; If a user has subclassed BlogEntry it should not be turned into a BlogPost.

PR #260, opened 2015-06-02
Escape $stage

micmania1:
Although I don't *think* this is vulnerable, the $stage variable ultimately comes from a $_GET parameter so should be escaped. Without this the security remains weak and could cause issues if the Versioned code is changed.

### DIFF
--- a/code/extensions/BlogFilter.php
+++ b/code/extensions/BlogFilter.php
@@ -23,6 +23,7 @@ class BlogFilter extends Lumberjack {
 				$stage = '_' . $stage;
 			}
 
+			$stage = Convert::raw2sql($stage);
 			$dataQuery = $staged->dataQuery()
 				->innerJoin('BlogPost', sprintf('"BlogPost%s"."ID" = "SiteTree%s"."ID"', $stage, $stage))
 				->where(sprintf('"PublishDate" < \'%s\'', Convert::raw2sql(SS_Datetime::now())));


tractorcow:
The problem is Convert::raw2sql doesn't correctly escape SQL identifiers, only sql values, so it's not really the correct way to escape this.

Also, it's safely isolated in Versioned::choose_site_stage which forces the value to Stage or Live only.

micmania1:
It's not safely isolated in Versioned::set_reading_mode() though. If a dev called Versioned::set_reading_mode($_GET['stage']) they would be vulnerable.

I wouldn't be confident in just removing double quotes - you could end the string with a backslash and break the query. That might not be an issue but I have no idea what else would be possible and i'm definitely not an expert on the subject.

Maybe the most simple & effective solution here is to ensure the value is either '_Live' or empty?

tractorcow:
If the user writes insecure code, it's not the responsibility of the framework to protect them. In your example the user has introduced their own vulnerability.

In any case, the code you've written would result in a broken query in the case of an injection attack. You'd just be mis-using the raw2sql method to make the query break in a different way.

```
$stage = $stage === 'Stage' ? '' : '_Live';
```

Would solve this issue without broken queries, without using the wrong api, and without injection vulnerabilities. You just can't have stages other than Stage or Live, but since you have decided not to trust any custom stage parameters, that's the limitation you're enforcing.

PR #277, opened 2015-06-27
FIX Migration of blog 1.0 to 2.0 was broken

camfindlay:
@assertchris @tractorcow I found the existing blog migration task actually broken my blog site when run (only local of course :D).

I have reworked it a little to ensure that those posts that were published or draft remain so in the new 2.0 version of blog.

Fixed the issue of tags being repeatedly migrated for each BlogEntry which caused lots of duplicate tags.

A few other minor hacks and improvements. It's not the prettiest but it works. Not I have removed the migration task from running in the build task. I figure it is something that should be consciously run by the migrator rather than slipping it in during a build task. This is of course up for debate however this is at least a step forward for blog 1.0 -> 2.0 migration.

### DIFF
--- a/code/compat/pages/BlogEntry.php
+++ b/code/compat/pages/BlogEntry.php
@@ -34,30 +34,53 @@ class BlogEntry extends BlogPost implements MigratableObject {
 	 * {@inheritdoc}
 	 */
 	public function up() {
+		
+		//Migrate comma separated tags into BlogTag objects.
 		foreach($this->TagNames() as $tag) {
-			if($this->Tags()->filter('Title', $tag)->count()) {
-				continue;
+			
+			$existingTag = BlogTag::get()->filter('Title', $tag);
+			if($existingTag->count()) {
+				//if tag already exists we will simply add it to this post.
+				$tagObject = $existingTag->First();
+			
+			} else {
+
+				//if the tag is now we create it and add it to this post.
+				$tagObject = new BlogTag();
+				$tagObject->Title = $tag;
+				$tagObject->BlogID = $this->ParentID;
+				$tagObject->write();
+
+				
 			}
 
-			$tagObject = new BlogTag();
-			$tagObject->Title = $tag;
-			$tagObject->BlogID = $this->ParentID;
-
-			$tagObject->write();
-
-			$this->Tags()->add($tagObject);
+			if($tagObject){
+				$this->Tags()->add($tagObject);
+			}
 		}
 
-		$this->PublishDate = $this->Date;
-		$this->AuthorNames = $this->Author;
-
+		//Store if the original entity was published or not (draft)
+		$published = $this->IsPublished();
 		// If a user has subclassed BlogEntry, it should not be turned into a BlogPost.
 		if($this->ClassName === 'BlogEntry') {
 			$this->ClassName = 'BlogPost';
 			$this->RecordClassName = 'BlogPost';
 		}
+		//Migrate these key data attributes
+		$this->PublishDate = $this->Date;
+		$this->AuthorNames = $this->Author;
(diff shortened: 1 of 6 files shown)


code/compat/pages/BlogHolder.php
```diff
@@ -31,14 +31,36 @@ public function canCreate($member = null) {
 		return false;
 	}
 
+
+	//Overload these to stop the Uncaught Exception: Object->__call(): the method 'parent' does not exist on 'BlogHolder' error.
+	public function validURLSegment() {
```
dhensby: Why are these required? Is it legacy? In a major upgrade your expect things to break and they should be in an upgrade guide

dhensby:
All looks good except my one comment

code/compat/pages/BlogHolder.php
camfindlay: Without these the migration fails with a Uncaught Exception: Object->__call(): the method 'parent' does not exist on 'BlogHolder' error. Perhaps there is some better way to handle this (I just don't intimately know that area of framework).

A search turned up this article from a while back, perhaps related? http://fvue.nl/wiki/SilverStripe:_Method_parent_does_not_exist_on_Page_in_sapphire/core/Object.php

I expect things to break in a major version however this is an attempt to help those on 1.0 migrate up to 2.0.

dhensby: Ah, this is in the compat code, that should be fine

camfindlay:
I will also update the docs to reflect the separate migration task.

code/compat/pages/BlogEntry.php
```diff
@@ -34,30 +34,53 @@ public function canCreate($member = null) {
...
 	 */
 	public function up() {
+		
+		//Migrate comma separated tags into BlogTag objects.
 		foreach($this->TagNames() as $tag) {
-			if($this->Tags()->filter('Title', $tag)->count()) {
-				continue;
+			
+			$existingTag = BlogTag::get()->filter('Title', $tag);
+			if($existingTag->count()) {
```
tractorcow: Oops, there is a problem here; You can't just search all tags everywhere; you need to filter against those tags against the current blog, since tags are blog specific.

With this code you can end up with a tag from one blog being assigned to a blogpost from another.

The original code was wrong in that it was selecting from `$this->Tags()` when it should be selecting from `$this->Parent()->Tags()` to select from the parent blog.

camfindlay:
Cheers @tractorcow will look at another patch shortly then (or you could just click edit and change this if you have the fix ;) ) Will $this->Parent even work in this context or could we simply grab the ParentID from the current BlogEntry/BlogPost?

tractorcow:
It might be safer to do `BlogTag::get()->filter(array('TItle' => $tag, 'BlogID' => $this->ParentID)`, which should work fine even if the parent page isn't migrated.

camfindlay:
Cheers, I'll find a moment and test locally before a PR.

code/compat/pages/BlogEntry.php
```diff
@@ -79,17 +102,6 @@ public function TagNames() {
 		return $results;
 	}
 
-	/**
-	 * {@inheritdoc}
-	 */
-	public function requireDefaultRecords() {
```
tractorcow: So we don't upgrade automatically when doing dev/build?

tractorcow: I only complain about this now, is because it means that we now need a new upgrade path for users. They can't just "install and rebuild" anymore, and to most people, they aren't used to any other process.

tractorcow:
I suggest we look at a `dev/upgrade` task in 3.3, since this is obviously not the right place to solve these issues.

PR #280, opened 2015-06-29
FIX Blog tag migration is per Blog

camfindlay:
Fix as mentioned on #277 by @tractorcow - tested locally against a 1.0 blog dataset.


tractorcow:
Thanks :D

Issue #281, opened 2015-06-29
BlogTagsWidget should have limit and sort


camfindlay:
Given tags are a folksonomy (taxonomy defined by the 'folks') there will likely be more entries that categories. User should be able to limit the number shown and order by either alphabetically or by number of times the tag is used (popularity?). Note blog 1.0 had this feature and the same should probably be applied to the BlogCategoriesWidget.

camfindlay:
:hand: high five

PR #286, opened 2015-07-08
Extended tags and categories widgets

assertchris:
Adds limit, sort and direction attributes to tags and categories widgets.

https://github.com/silverstripe/silverstripe-blog/issues/281

screen shot 2015-07-08 at 1 07 52 pm

screen shot 2015-07-08 at 1 07 59 pm

### DIFF
--- a/code/widgets/BlogCategoriesWidget.php
+++ b/code/widgets/BlogCategoriesWidget.php
@@ -26,7 +26,11 @@ class BlogCategoriesWidget extends Widget {
 	/**
 	 * @var array
 	 */
-	private static $db = array();
+	private static $db = array(
+		'Limit' => 'Int',
+		'Order' => 'Varchar',
+		'Direction' => 'Varchar',
+	);
 
 	/**
 	 * @var array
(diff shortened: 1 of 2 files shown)


code/widgets/BlogCategoriesWidget.php
```diff
@@ -39,13 +43,26 @@ class BlogCategoriesWidget extends Widget {
...
 	public function getCMSFields() {
-		$this->beforeUpdateCMSFields(function ($fields) {
-			/**
-			 * @var FieldList $fields
-			 */
-			$fields->push(
-				DropdownField::create('BlogID', _t('BlogCategoriesWidget.Blog', 'Blog'), Blog::get()->map())
+		$this->beforeUpdateCMSFields(function (FieldList $fields) {
+			$fields[] = DropdownField::create(
+				'BlogID', _t('BlogCategoriesWidget.Blog', 'Blog'), Blog::get()->map()
```
tractorcow: I love learning that this works. :D

code/widgets/BlogTagsWidget.php
```diff
@@ -55,11 +72,17 @@ public function getCMSFields() {
...
+		if (!$blog) {
+			return array();
 		}
 
-		return array();
+		$data = $this->data();
+
+		return $blog->Tags()
+			->limit($data->Limit)
+			->sort($data->Order, $data->Direction);
```
tractorcow: There's no SQL injection prevention here. :D

I would have a check to ensure 

```
if(!singleton('BlogCategory')->hasField($this->Order)) throw new ValidationException()
```

Or something. Probably on BlogTagsWidget::validate?

code/widgets/BlogTagsWidget.php
```diff
@@ -55,11 +72,17 @@ public function getCMSFields() {
...
-		if($blog = $this->Blog()) {
-			return $blog->Tags();
+		$blog = $this->Blog();
+
+		if (!$blog) {
+			return array();
 		}
 
-		return array();
+		$data = $this->data();
```
tractorcow: `data()` just returns `$this` so probably isn't necessary here.

code/widgets/BlogTagsWidget.php
```diff
@@ -55,11 +72,17 @@ public function getCMSFields() {
 	 * @return array
```
tractorcow: It's not actually returning an array; It's a Datalist isn't it?

PR #297, opened 2015-09-17
Allow injection on date

lekoala:
This change allow the use of the injector to inject another class for the date. I needed this because FormatI18N was not working as expected to localize the dates. I know it's suppose to work, but since there is no downside in allow injection, why not?


tractorcow:
Sorry I didn't get to your pull request in time, but can you please rebase this?

lekoala:
done !

PR #298, opened 2015-09-24
Update composer constraints for tagfield

kinglozzer:
See #293

PR #299, opened 2015-10-01
Use the tilde convention as per Module Standard.

camfindlay:


composer.json
```diff
@@ -8,7 +8,7 @@
 	],
 	"type": "silverstripe-module",
 	"require": {
-		"silverstripe/cms": ">=3.1.0",
+		"silverstripe/cms": "~3.1",
```
micmania1: Would it not be better to use `^`? What was the reason `~` was chosen as the standard?

micmania1: For reference: https://getcomposer.org/doc/articles/versions.md#caret

assertchris: I would prefer `^` over `~` because it's more inline with modern composer/npm version specification...

assertchris: I am fairly sure the module standard doesn't say use `~` over `^`. If it does, that wasn't added with group consensus... :D

kinglozzer:
@camfindlay could you incorporate/cherry-pick #298 into this too? :)

PR #301, opened 2015-10-13
Bugfix: avoid calling "new PaginatedList(null)"

JorisDebonnet:
/rss on a blog currently throws an error because PaginatedList gets called in the chain (somewhere within getMetaTitle) with apparently null as parameter. So if `$this->blogPosts` is null, don't go into that chain.

### DIFF
--- a/code/model/Blog.php
+++ b/code/model/Blog.php
@@ -962,7 +962,7 @@ class Blog_Controller extends Page_Controller {
 		 */
 		$dataRecord = $this->dataRecord;
 
-		$posts = new PaginatedList($this->blogPosts);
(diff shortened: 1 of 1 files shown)


micmania1:
I'm not sure this is the right place to fix it. Changing $this->blogPosts to $this->getBlogPosts() should mean null is never used in PaginatedList: https://github.com/JorisDebonnet/silverstripe-blog/blob/patch-1/code/model/Blog.php#L965

JorisDebonnet:
You're right, that's a better fix! Implemented now.

micmania1:
Thanks!

Issue #289, opened 2015-08-05
Sort Order of Blog Posts in Admin


basepath:
I haven't been able to find an alternative way of doing this so can the default sort be based on Publish Date? 

ie. in BlogPost.php add inner join as per below...
$pages = SiteTree::get()->innerJoin('BlogPost', '"SiteTree"."ID" = "BlogPost"."ID"')->filter(array(
	'ParentID' => $this->owner->ID,
	'ClassName' => $excluded
))->sort('"BlogPost"."PublishDate" DESC');

JorisDebonnet:
I also had issues with the ordering, and finally found where it actually happens: not in `BlogPost.php`, but in `BlogFilter.php`. 

```php
			$pages = BlogPost::get()->filter(array(
				'ParentID' => $this->owner->ID,
				'ClassName' => $excluded
			))->sort('"PublishDate" DESC');
```
suffices in `updateCMSFields` there, instead of the current situation which does `->sort('"SiteTree"."Created" DESC');`.

JorisDebonnet:
This is now fixed in the most recent commit! You should see it if you pull in the lastest dev version. So I think this can be closed.

PR #302, opened 2015-10-13
Sort BlogPosts in the CMS by PublishDate

JorisDebonnet:
I don't understand why you'd want to sort by CreationDate rather than PublishDate. It feels quite counterintuitive to have the list not sorted chronologically. 

I might be missing something, but if I'm not, this is an easy fix.

### DIFF
--- a/code/extensions/BlogFilter.php
+++ b/code/extensions/BlogFilter.php
@@ -71,10 +71,10 @@ class BlogFilter extends Lumberjack {
 		$excluded = $this->owner->getExcludedSiteTreeClassNames();
 
 		if(!empty($excluded)) {
-			$pages = SiteTree::get()->filter(array(
+			$pages = BlogPost::get()->filter(array(
 				'ParentID' => $this->owner->ID,
 				'ClassName' => $excluded
-			))->sort('"SiteTree"."Created" DESC');
+			));
 
 			$gridField = new BlogFilter_GridField(
 				'ChildPages',
--- a/code/model/BlogPost.php
+++ b/code/model/BlogPost.php
@@ -84,9 +84,11 @@ class BlogPost extends Page {
 	private static $allowed_children = array();
 
 	/**
+	 * The default sorting lists BlogPosts with an empty PublishDate at the top.
+	 * 
 	 * @var string
 	 */
-	private static $default_sort = 'PublishDate DESC';
+	private static $default_sort = '"PublishDate" IS NULL DESC, "PublishDate" DESC';
 
 	/**
 	 * @var bool
@@ -211,8 +213,10 @@ class BlogPost extends Page {
 				_t('BlogPost.AdditionalCredits', 'Additional Credits'),
 				null,
 				1024
+			)->setDescription(_t(
+					'BlogPost.PublishDate_Description',
+					'If some authors of this post don\'t have CMS access, enter their name(s) here. You can separate multiple names with a comma.')
 			);
-			$authorNames->setDescription('If some authors of this post don\'t have CMS access, enter their name(s) here. You can separate multiple names with a comma.');
 
 			if(!$self->canEditAuthors()) {
 				$authorField = $authorField->performDisabledTransformation();
@@ -221,6 +225,12 @@ class BlogPost extends Page {
 
 			$publishDate = DatetimeField::create('PublishDate', _t('BlogPost.PublishDate', 'Publish Date'));
 			$publishDate->getDateField()->setConfig('showcalendar', true);
+			if(!$self->PublishDate) {
+				$publishDate->setDescription(_t(
+						'BlogPost.PublishDate_Description',
+						'Will be set to "now" if published without a value.')
+				);
+			}
 
 			// Get categories and tags
 			$parent = $self->Parent();
@@ -365,8 +375,7 @@ class BlogPost extends Page {
 	/**
 	 * {@inheritdoc}
 	 *
-	 * Update the PublishDate to now, if being published for the first time, and the date hasn't
-	 * been set to the future.
+	 * Update the PublishDate to now if the BlogPost would otherwise be published without a date.
 	 */
 	public function onBeforePublish() {
 		/**
@@ -374,7 +383,7 @@ class BlogPost extends Page {
 		 */
 		$publishDate = $this->dbObject('PublishDate');
 
-		if($publishDate->InPast() && !$this->isPublished()) {
+		if(!$publishDate->getValue()) {
 			$this->PublishDate = SS_Datetime::now()->getValue();
 			$this->write();
 		}
@@ -414,7 +423,7 @@ class BlogPost extends Page {
 		if(!parent::canView($member)) {
 			return false;
 		}
-		
+
 		/**
 		 * @var SS_Datetime $publishDate
 		 */
@@ -624,10 +633,6 @@ class BlogPost extends Page {
 	 */
 	protected function onBeforeWrite() {
 		parent::onBeforeWrite();
-
-		if(!$this->PublishDate) {
-			$this->PublishDate = SS_Datetime::now()->getValue();
-		}
 
 		if(!$this->exists() && ($member = Member::currentUser())) {
 			$this->Authors()->add($member);


JorisDebonnet:
(also see #289)

assertchris:
Trouble with using `BlogPost` directly is it ignores subclasses. I think that's why the `SiteTree` reference is there...

JorisDebonnet:
Are those classes that extend `BlogPost` then? Won't they automatically be included if you list all BlogPosts?

micmania1:
I'm pretty sure excluded pages don't have to be subclasses of BlogPost which means PublishDate might not always exist. However, that would usually be an edge-case.

tractorcow:
Probably because some posts aren't published, and it makes sorting unpredictable. Every page has a created date and that doesn't change.

stevie-mayhew:
@tractorcow every blog post does have a publish date: https://github.com/silverstripe/silverstripe-blog/blob/master/code/model/BlogPost.php#L368

They also have a default sort set as `PublishDate DESC` and I don't think it makes sense to override it in the administration as it provides a different experience from the front end which can be confusing for editors: https://github.com/silverstripe/silverstripe-blog/blob/master/code/model/BlogPost.php#L89

tractorcow:
@stevie-mayhew that method is only called if the page is published. If it's unpublished and still in draft, that code will never run.

JorisDebonnet:
Well in any case, when I test it, a never-published BlogPost does have a PublishDate. But actually when you look at the form, why wouldn't it? It's a value you fill in yourself, so it's saved, whether the post is in fact published already or not. 

So I think the only issue with this is if any children of a Blog page are not of the type BlogPost, and simply don't have a PublishDate. But do we really want that?... I guess in that case it could sort by PublishDate if it can find it, and otherwise by creation date (although I can only imagine some ugly SQL to achieve that, namely a LEFT JOIN and a COALESCE in the ORDER).

jonom:
I agree with @stevie-mayhew that the gridfield should respect the default sort... but I also think that draft posts should be highlighted and they would be pushed to the bottom of the list if we did that. Here's a possible solution: https://github.com/silverstripe/silverstripe-blog/issues/305#issuecomment-149608790

tractorcow:
How about

`ORDER BY "PublishDate" IS NULL DESC, "PublishDate" DESC`?

jonom:
@tractorcow That is a cool idea.

stevie-mayhew:
Nice @tractorcow 

@JorisDebonnet do you want to implement that change in this PR?

assertchris:
Does not handle tables missing PublishedDate fields. Is that ok?

jonom:
@assertchris If I've read the code right, out of the box the only allowed children for `Blog` is `BlogPost` so shouldn't run in to this problem I believe as PublishDate will always exist on the table of pages in the gridfield. If people do override `Blog::$allowed_children` to allow page types that aren't subclasses of `BlogPost` I think you'd only have a problem if you also set `$show_in_sitetree = false` on those additional page types, as this is needed for lumberjack to cut the page out of the sitetree and include it in the gridfield. Doing `BlogPost::get()` instead of `SiteTree::get()` as proposed in this PR should mean we never run in to a case where the `PublishDate` field is missing for page types included in the gridfield. Does come with the possible downside that regular pages can't be shown in the gridifled - but they should still be available in the sitetree so this seems suitable to me - I think it would be unreasonable to expect pages that aren't blog posts to show up in a gridfield of blog posts.

JorisDebonnet:
Making sure Drafts stay in view makes sense to me, but ordering by `"PublishDate" IS NULL` won't make that happen: any BlogPost, including drafts and any subclasses, have a PublishDate. Even removing the field from the CMS form does not make it NULL -- I tried it. 

It's because [onBeforeWrite](https://github.com/silverstripe/silverstripe-blog/blob/master/code/model/BlogPost.php#L629) (just like `onBeforePublish`) gives PublishDate a default value. Which makes total sense in my opinion: PublishDate does not say anything about whether a BlogPost is published; it is nothing more than an information field to be sorted by and displayed on the website. 

In that sense, I don't think it's wrong to quite simply sort by PublishDate, as the PR does now. Drafts will still be visible, assuming you don't manually put its PublishDate way in the past. 

The only remaning consideration then is whether it's a problem that non-BlogPost children won't be listed in the GridField. I don't think it is, for the same reason @jonom mentioned above.

jonom:
Hmmm... there's some stuff going on here that doesn't make sense to me

1. When a post is created for the first time, the publish date is set to the current date/time (in onBeforeWrite)
2. When a post is published for the first time, that date is fast-forwarded to the current date/time if it's in the past (in onBeforePublish)

This seems unnecessarily restrictive to me - what if you legitimately want to set a blog post to a date in the past? It's going to be updated on publish and confuse the CMS user.

I think new blog posts should start with a NULL PublishDate, and we should only update this when the page is published for the first time, *if* it's still empty. The PublishDate field in the CMS could display a hint like "If left blank, will be auto-filled when published".

JorisDebonnet:
I agree on your points. I had not even realized yet that Drafts with a date in the past (e.g. 1 October) are updated behind the scenes when you publish them, that certainly shouldn't happen! 

I would say the PublishDate line in onBeforeWrite should be removed, and the onBeforePublish should do what onBeforeWrite currently does (i.e. default the PublishDate to "now" if it wasn't set yet). 

And then we can indeed sort by `"PublishDate" IS NULL DESC, "PublishDate" DESC`!

Shall I put these changes in this PR?

jonom:
I vote for yes, put them in the PR :) can you make sure to leave a `!$this->isPublished()` check in onBeforePublish so it only happens the first time you publish, otherwise yes it would just be the stuff from onBeforeWrite.

jonom:
Maybe include a code comment with the sort order too explaining the rationale so it's easier to understand its purpose in future :)

JorisDebonnet:
I guess the "set `PublishDate = now` if empty" only *needs* to happen on first publish, but I don't think it hurts if it happens on subsequent publications. It only happens if it's still NULL after all, so that's unlikely to happen more than once. But if a published BlogPost does somehow end up with a NULL PublishDate,... I would say it wouldn't hurt to sanitize it to "now" on the next publish/save! (otherwise it ends up being buried in the CMS ordering anyway)

jonom:
Yeah I guess that's true. Let's just make sure the CMS field description indicates that behaviour (if blank will be filled) so there are no surprises.

code/model/BlogPost.php
```diff
@@ -211,8 +211,10 @@ public function getCMSFields() {
 				_t('BlogPost.AdditionalCredits', 'Additional Credits'),
 				null,
 				1024
+			)->setDescription(_t(
+					'BlogPost.PublishDate_Description',
+					'If some authors of this post don\'t have CMS access, enter their name(s) here. You can separate multiple names with a comma.')
 			);
-			$authorNames->setDescription('If some authors of this post don\'t have CMS access, enter their name(s) here. You can separate multiple names with a comma.');
```
JorisDebonnet: (I added i18n support here just for completeness sake, although this line has nothing to do with this PR)

JorisDebonnet:
Well: updated.

code/extensions/BlogFilter.php
```diff
@@ -71,10 +71,11 @@ public function updateCMSFields(FieldList $fields) {
 		$excluded = $this->owner->getExcludedSiteTreeClassNames();
 
 		if(!empty($excluded)) {
-			$pages = SiteTree::get()->filter(array(
+			$pages = BlogPost::get()->filter(array(
 				'ParentID' => $this->owner->ID,
 				'ClassName' => $excluded
-			))->sort('"SiteTree"."Created" DESC');
```
jonom: Can you remove this `sort` method call so the default sort is respected, and update the default sort on BlogPost instead?

JorisDebonnet: Ah I see. Done!

code/model/BlogPost.php
```diff
@@ -221,6 +223,12 @@ public function getCMSFields() {
 
 			$publishDate = DatetimeField::create('PublishDate', _t('BlogPost.PublishDate', 'Publish Date'));
 			$publishDate->getDateField()->setConfig('showcalendar', true);
+			if(!$self->PublishDate) {
```
jonom: This is probably okay in this case as it covers the most significant 'blank when created' scenario, but I tend to avoid bringing server-side logic in to field descriptions as we don't know what's going to happen client side. If the PublishDate field has a value initially and someone deletes it they won't receive a warning that it's going to be auto-filled. Ideally we would always include the description and use javascript for conditional display.

jonom:
Looks good to me! Is it working as expected?

JorisDebonnet:
I implemented it in the application where I needed this, and it does work as expected there.

tractorcow:
Weren't you going to add a `default_sort` to BlogPost as well?

JorisDebonnet:
Oh, strange, looks like it wasn't picked up in my commit!.. Included now.

PR #310, opened 2015-10-29
FIX If grant_user_access is set then return filtered list

camfindlay:
I think there is a bug here in that a config value doesn't actually filter the returned Members. This causes a problem on large sites with 1000's of users as it returns them all and you often run into a out of memory problem. I should be able to set a permission value which qualifies a list of appropriate users that can manage blog roles.

Is it as simple as reversing the conditional?

### DIFF
--- a/code/model/Blog.php
+++ b/code/model/Blog.php
@@ -360,11 +360,11 @@ class Blog extends Page implements PermissionProvider {
 	 */
 	protected function getCandidateUsers() {
 		if($this->config()->grant_user_access) {
-			return Member::get();
-		} else {
 			return Permission::get_members_by_permission(
 				$this->config()->grant_user_permission
 			);
+		} else {
+			return Member::get();
 		}
 	}
 


camfindlay:
I'm looking to start working on some enhancements for this use case so that we have support for large userbase sites.

camfindlay:
Can we agree in principle this could be an issue @tractorcow ?

tractorcow:
Hi @camfindlay, thanks for looking into this, but the existing code is correct.

The reason for returning all users if "grant_user_permissions" is enabled is because users are not REQUIRED to have the permission, because if they don't have it, they will be granted it later.

grant_user_permissions means "they will be granted if they don't have it" not "they must have been granted these prior".

Why don't you create a permission via the PermissionProvider interface somewhere, assign it to the groups you want to have blog access, and set the config to this permission?

tractorcow:
I'd also be open to having an extension point on `getCandidateUsers` to let you customise this list.

tractorcow:
Possibly the other `grant_user_permission` is mis-named. It could probably be split into `required_user_permission` and `grant_user_permission`; One for this filter and the other in the code that  actually promotes the user.

camfindlay:
Perhaps the better fix is to be able to set an optional user group to restritct the pool of blog users to select from at per #311 ?

camfindlay:
Will be adding an extension here and another other in BlogPost, see #311 for updates and eventual pull request.

Issue #311, opened 2015-10-29
BUG Sites with large number of users (Member) breaks


camfindlay:
There are a number of fields that list Member to be selected to assign authors etc. These seem to all use Member::get() which given a sizable website with a large number of users brings everything to a grinding halt.

There should be some way to qualify a Member for select rather than a blanket "get all the things" for these situations.

Example in BlogPost:
```php
$authorField = ListboxField::create(
				'Authors',
				_t('BlogPost.Authors', 'Authors'),
				Member::get()->map()->toArray()
			)->setMultiple(true);
```

This kills my large site when I drill into a BlogPost, however if I add a simple ->limit(10) to the Member ORM call, no problem, page loads (though it's not useful in this state).

I'd want to be able to set a config value denoting a Security group I can add any users to which becomes my pool of potential authors that are selectable. This keeps the whole thing more controllable and avoids issues with large user bases.

micmania1:
@camfindlay Can this be closed?

camfindlay:
yip sure can! #312 fixes and we've tested on production in ss.org :+1:

PR #312, opened 2015-10-30
ENHANCEMENT filtering for large user base sites.

camfindlay:
Improvements for #311

### DIFF
--- a/code/model/BlogPost.php
+++ b/code/model/BlogPost.php
@@ -16,6 +16,14 @@
  * @property int $ParentID
  */
 class BlogPost extends Page {
+
+	/**
+	 * Same as above, but for list of users that can be
+	 * given credit in the author field for blog posts
+	 * @var string|bool false or group code
+	 */
+	private static $restrict_authors_to_group = false;
+
 	/**
 	 * @var array
 	 */
@@ -205,7 +213,7 @@ class BlogPost extends Page {
 			$authorField = ListboxField::create(
 				'Authors',
 				_t('BlogPost.Authors', 'Authors'),
-				Member::get()->map()->toArray()
+				$this->getCandidateAuthors()->map()->toArray()
 			)->setMultiple(true);
 
 			$authorNames = TextField::create(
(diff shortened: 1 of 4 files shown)


camfindlay:
Fixing test fail now.

camfindlay:
Unsure why test here is failing, would appreciate some help on this (code is working as expected however good to have tests in place).

tractorcow:
Because you are calling a protected method from outside the class. :D

camfindlay:
@tractorcow in that case should I simply open the visibility of the method?

camfindlay:
Updated both the getCandidate___ methods to be public, I can see as these are getMethods they should be accessible on the view layer and hence should both be public methods.

tractorcow:
@camfindlay unfortunately that's a semver violation for a non-major release; We can't change method visibility of existing functions without releasing blog 3.0. =/

If you need to test it, have a subclass that exposes those methods just for testing. I.e. have a spy subclass that implements `TestOnly`.

camfindlay:
If I put getCandidateUser back to protected that should be fine then right? getCandidateAuthors is a new method and not breaking any API? Minor version release?

camfindlay:
@tractorcow I've put the method on Blog back to previous visibility. Would this now be a BC PR?

tractorcow:
Looks good mate. :D

code/model/BlogPost.php
```diff
@@ -205,7 +213,7 @@ public function getCMSFields() {
 			$authorField = ListboxField::create(
 				'Authors',
 				_t('BlogPost.Authors', 'Authors'),
-				Member::get()->map()->toArray()
+				$this->getCandidateAuthors()->map()->toArray()
```
tractorcow: Oops, turned out this broke compatibility with php 5.3 since `$this` is used in a closure. :)

dhensby:
To test protected methods you should use ReflectionMethod. It's been done a few times in the core tests if you need an example

tractorcow:
True, but should unit tests be covering protected methods, as a matter of practice? I'd rather that they were publicly defined API.

If they should be unit tested, and thus have predictable behaviour (that _should_ be predictable) then it should follow the same restrictions that semver place on public API. Otherwise, internal refactoring will mean constantly having to change the assumptions of tests, even when no API has actually changed. I had this trouble with the new `AssetStore` and how it broke any tests that had file fixtures.

But I digress... we are getting into philosophical grounds here. :D

dhensby:
Unit tests are to check units of functionality work as expected. If x goes in y comes out.

If the method is changed that now z comes out, well, tests need to be updated (as they would with public methods).

Tests aren't there just to check the public API works, they're there to check there aren't regressions at all.

tractorcow:
Then it feels like we are treating protected methods as public API, so it may as well be public.

dhensby:
except semver

And by that arguement, every function should be public, which isn't the case, and we just kind of accept regressions to protected or private methods as course of life, which seems silly.

If someone wants to re-factor the internals of a function to be more efficient we want to know x -> y and not x -> z; that's what tests are for...

tractorcow:
I think we simply have a difference in opinion in testing goals. :)

I'm going to have a think about this and come back to this discussion later.

PR #314, opened 2015-10-30
BUGFIX: Blogs could not be nested under anything other than a Blog

UndefinedOffset:
Due to how the [Hierarchy](https://github.com/silverstripe/silverstripe-framework/blob/3.2/model/Hierarchy.php#L41-L45) class adds the has_one relationship to the parent when the blog module's BlogFilter extension is added it overrides the has_one relationship pointing it to another Blog page. Which causes the blog to be considered orphaned which in turn is preventing non-authenticated/non-cms users from being able to view the blog and it's decedents. This pull requests addresses the issue by overriding Hierarchy::get_extra_config() and forcing the Parent has_one to be set to SiteTree.

### DIFF
--- a/code/extensions/BlogFilter.php
+++ b/code/extensions/BlogFilter.php
@@ -8,6 +8,12 @@
  * @subpackage blog
  */
 class BlogFilter extends Lumberjack {
+	public static function get_extra_config($class, $extension, $args) {
+		return array(
+			'has_one' => array('Parent' => 'SiteTree')
+		);
+	}
+	
 	/**
 	 * {@inheritdoc}
 	 */


UndefinedOffset:
I guess the other way to do this may to simply be just have BlogFilter::get_extra_config() do nothing. Since Hierarchy::get_extra_config() has already do that.

tractorcow:
Doesn't this risk creating a ParentID database field on the BlogPost table, rather than on the SiteTree table?

tractorcow:
Oh, I see the problem.

It's because LumberJack extends Hierarchy.

If you change your code to the below, does it fix your problem?

```php
public static function get_extra_config($class, $extension, $args) {
	return array();
}
```

tractorcow:
Fortunately it looks like moving pages to different parents somehow writes the value to both tables, so removing the Blog.ParentID shouldn't hurt.

tractorcow:
Fix at https://github.com/silverstripe/silverstripe-lumberjack/pull/13

UndefinedOffset:
Ya that works @tractorcow much better solution than this hack, I think this could be closed

tractorcow:
Well, when someone merges the lumberjack fix, sure.

stevie-mayhew:
This is still an issue as there hasn't been a release for Lumberjack with the fix yet :smile: https://github.com/silverstripe/silverstripe-lumberjack/commit/bf0db23c1d606ca62fef87ee038223e35f751151#commitcomment-14406875

PR #315, opened 2015-10-31
ENHANCEMENT Refactor of blog migration task to be more reliable.

camfindlay:
Previous version of the migration task often cased site breakages. This goes back to basics, moving the data with SQL calls and limited ORM use to ease the data is not modified when mapped to the new blog 2.0 data structure.

This has been tested on a large blog 1.0 dataset.


code/tasks/BlogMigrationTask.php
```diff
@@ -0,0 +1,363 @@
...
+		try {
+			DB::query('SELECT ID FROM Blog');
+			DB::query('SELECT ID FROM BlogPost');
+			echo "Blog and BlogPost tables exist, you are good to migrate" . $this->eol;
+		} catch (Exception $e) {
+			echo 'Ensure you have run a <a href="/dev/build" target="_blank">dev/build</a>' . $this->eol;
+		}
+
+		//THE MIGRATION
+		if(!$request->getVar('migration') == 1){
```
tractorcow: `!(var) == 1`?

code/tasks/BlogMigrationTask.php
```diff
@@ -0,0 +1,363 @@
...
+			exit;
+		}
+
+		$this->extend('onBeforeBlogMigration',$request, $this->eol);
+
+		//BlogPost Migration
+		//Migrate BlogEntry to BlogPost include _Live and _versions
+		try {
+			DB::query('
+              INSERT INTO BlogPost (ID, PublishDate, AuthorNames)
```
tractorcow: This SQL won't work in postgres; Can you please double-quote all columns / table names?

code/tasks/BlogMigrationTask.php
camfindlay: typo will fix

code/tasks/BlogMigrationTask.php
camfindlay: If that is the case can you please add Postgres into the travis config please if we are going to require official testing against this DB.

camfindlay:
Can you check my updates @tractorcow ?

code/tasks/BlogMigrationTask.php
camfindlay: actually, not a typo... I am testing that the query string 'migration' includes the value 1. Migration is something you want to be double sure that you want to actually run.

camfindlay:
Updated both the getCandidate___ methods to be public, I can see as these are getMethods they should be accessible on the view layer and hence should both be public methods.

code/tasks/BlogMigrationTask.php
micmania1: why the exclamation mark at the start?

camfindlay: This provides a few options and checks to people before the migration runs. It also runs a check to see if the database structure exists (due to a dev/build). If it doesn't it kills the migration which we dont want to happen (it would be a bad experience for people migrating).

tractorcow: `if(!$var == 'value')` means `$var2 = !$var; if($var2 == 'value')`.

boolean not `!` operator has higher precedence than the loose-comparison one `==`

What you mean is `if($request->getVar('migration') != 1)`

camfindlay: Yip correct, I'll update...

camfindlay:
Need to add some documentation around the extension points here.

dhensby:
Ping @tractorcow

code/tasks/BlogMigrationTask.php
```diff
@@ -0,0 +1,357 @@
...
+	 * @param object $request
+	 */
+	public function run($request) {
+		// end of line output depending on CLI or browser run.
+		$this->eol = Director::is_cli() ? PHP_EOL : "<br>";
+
+		//PRE-FLIGHT CHECK
+		// Ensure a dev build has been run by check for some expected tables.
+		try {
+			DB::query('SELECT "ID" FROM "Blog"');
```
tractorcow: Please use `ClassInfo::hasTable('Blog')` etc rather than relying on error handling to report these tables.

code/tasks/BlogMigrationTask.php
```diff
@@ -0,0 +1,357 @@
...
+			DB::query('SELECT "ID" FROM "BlogPost"');
+			echo "Blog and BlogPost tables exist, you are good to migrate" . $this->eol;
+		} catch (Exception $e) {
+			echo 'Ensure you have run a <a href="/dev/build" target="_blank">dev/build</a>' . $this->eol;
```
tractorcow: We should `return;` if the migration can no longer continue.

code/tasks/BlogMigrationTask.php
tractorcow: You'll also need another error if `BlogEntry` and `BlogHolder` don't exist, since there will be nothing to migrate FROM.

code/tasks/BlogMigrationTask.php
```diff
@@ -0,0 +1,357 @@
...
+			exit;
+		}
+
+		$this->extend('onBeforeBlogMigration',$request, $this->eol);
+
+		//BlogPost Migration
+		//Migrate BlogEntry to BlogPost include _Live and _versions
+		try {
+			DB::query('
+              INSERT INTO "BlogPost" ("ID", "PublishDate", "AuthorNames")
```
tractorcow: Might be cleaner to use HEREDOC, or at least fix the indentation here. ;)

code/tasks/BlogMigrationTask.php
```diff
@@ -0,0 +1,357 @@
...
+              (
+              	SELECT "ID", "Date", "Author"
+              	FROM "BlogEntry"
```
tractorcow: WHERE `WHERE "ID" NOT IN (SELECT "BlogPost"."ID" FROM "BlogPost")` to prevent errors, and allow incremental updates. :D

Maybe your DB::query could report the number of records copied, with DB::affectedRows()?

code/tasks/BlogMigrationTask.php
```diff
@@ -0,0 +1,357 @@
...
+              )
+            ');
+			echo "Migrated BlogEntry to BlogPost" . $this->eol;
+		} catch (Exception $e) {
+			echo "BlogEntry to BlogPost migration already run, moving along..." . $this->eol;
+		}
+		//BlogPost_Live
+		try {
+			DB::query('
+              INSERT INTO "BlogPost_Live" ("ID", "PublishDate", "AuthorNames")
```
tractorcow: Looks like code duplication. Maybe make $stage a variable that's passed to a method that does the migration for this table?

code/tasks/BlogMigrationTask.php
```diff
@@ -0,0 +1,357 @@
...
+		} catch (Exception $e) {
+			echo "BlogEntry_Live to BlogPost_Live migration already run, moving along..." . $this->eol;
+		}
+		//BlogPost_version
+		try {
+			DB::query('
+              INSERT INTO "BlogPost_versions" ("ID", "RecordID", "Version", "PublishDate", "AuthorNames")
+              (
+              	SELECT "ID", "RecordID", "Version", "Date", "Author"
+              	FROM "BlogEntry_versions"
```
tractorcow: The same `WHERE "ID" NOT IN (SELECT "BlogPost_versions"."ID" FROM "BlogPost_versions")` As above would probably be fine, assuming that the version ID uniquely identifies the same RecordID / Version pair. :)

code/tasks/BlogMigrationTask.php
```diff
@@ -0,0 +1,357 @@
...
+			echo "BlogEntry_versions to BlogPost_versions migration already run, moving along..." . $this->eol;
+		}
+		//SiteTree ClassName BlogEntry to BlogPost
+		try {
+			DB::query('UPDATE "SiteTree" SET "ClassName" = \'BlogPost\' WHERE "ClassName" = \'BlogEntry\'');
+			DB::query('UPDATE "SiteTree_Live" SET "ClassName" = \'BlogPost\' WHERE "ClassName" = \'BlogEntry\'');
+			DB::query('UPDATE "SiteTree_versions" SET "ClassName" = \'BlogPost\' WHERE "ClassName" = \'BlogEntry\'');
+			echo "Updated ClassName reference to BlogPost" . $this->eol;
+		} catch (Exception $e) {
+			echo $e;
```
tractorcow: We shouldn't echo here.

code/tasks/BlogMigrationTask.php
```diff
@@ -0,0 +1,357 @@
...
+			echo "SiteTree BlogPost ClassName migration already run, moving along..." . $this->eol;
```
tractorcow: I don't think these queries should be failing in any case.

code/tasks/BlogMigrationTask.php
```diff
@@ -0,0 +1,357 @@
...
+			echo "Migrated BlogHolder to Blog" . $this->eol;
+		} catch (Exception $e) {
+			echo "BlogHolder to Blog migration already run, moving along..." . $this->eol;
+		}
+		//Blog_Live
+		try {
+			DB::query('
+              INSERT INTO "Blog_Live" ("ID", "PostsPerPage")
+              (
+              	SELECT "BlogHolder_Live"."ID", 10
```
tractorcow: We should be setting the default for this value at a DB level.

```php
/**
     * @var array
     */
    private static $db = array(
        'PostsPerPage' => 'Int(10)',
    );
```

code/tasks/BlogMigrationTask.php
```diff
@@ -0,0 +1,357 @@
...
+	}
+
+	/*
+     * Migrate the tags
+	 *
+	 * @return int number of migrated tags
+     */
+	protected function migrateTags(){
+		//1. Get all BlogEntry ID and comma separated tags into an array
+		$blogtags = DB::query('SELECT ID, Tags FROM BlogEntry_Live')->map('ID','Tags');
```
tractorcow: Please use ansi-sql quotes here. `SELECT "ID", "Tags"...` etc.

code/tasks/BlogMigrationTask.php
```diff
@@ -0,0 +1,357 @@
...
+		echo "Migrated TagCloudWidget to BlogTagsWidget" . $this->eol;
+	}
+
+	/*
+     * Clean up old tables
+	 *
+	 * @return boolean true of the clean up ran without an issue.
+     */
+	protected function cleanUp() {
+		DB::query('DROP TABLE "BlogEntry"');
```
tractorcow: Should we back these up to `_obsolete_BlogEntry` tables instead?

code/tasks/BlogMigrationTask.php
```diff
@@ -0,0 +1,357 @@
...
+			echo "BlogHolder to Blog migration already run, moving along..." . $this->eol;
+		}
+		//SiteTree ClassName BlogEntry to BlogPost
+		try {
+			DB::query('UPDATE "SiteTree" SET "ClassName" = \'Blog\' WHERE "ClassName" = \'BlogHolder\'');
+			DB::query('UPDATE "SiteTree_Live" SET "ClassName" = \'Blog\' WHERE "ClassName" = \'BlogHolder\'');
+			DB::query('UPDATE "SiteTree_versions" SET "ClassName" = \'Blog\' WHERE "ClassName" = \'BlogHolder\'');
+			echo "Updated ClassName reference to Blog" . $this->eol;
+		} catch (Exception $e) {
+			echo $e;
```
tractorcow: Same dealio as per the BlogPost migration comments I left above. ;)

code/tasks/BlogMigrationTask.php
```diff
@@ -0,0 +1,357 @@
...
+		}
+
+		//Migrate BlogTree to Page
+		try {
+			DB::query('UPDATE "SiteTree" SET "ClassName" = \'Page\' WHERE "ClassName" = \'SiteTree\'');
```
tractorcow: I'm not sure that a blog migrator should be doing this.

code/tasks/BlogMigrationTask.php
```diff
@@ -0,0 +1,357 @@
...
+     */
+	protected function migrateTags(){
+		//1. Get all BlogEntry ID and comma separated tags into an array
+		$blogtags = DB::query('SELECT ID, Tags FROM BlogEntry_Live')->map('ID','Tags');
+		$tagcount = 0;
+		//2. Foreach split the tags into own array
+		foreach($blogtags as $blogpostid => $tags){
+			foreach($this->tagNames($tags) as $tag) {
+				//3a. If it's an existing tag, connect the BlogPost and BlogTag as many_many
+				$existingTagID = DB::query('SELECT ID FROM BlogTag WHERE Title = \'' .$tag. '\'')->value();
```
tractorcow: You will need to scope your query to only BlogTag items that belong to the parent Blog. Add a BlogTag.BlogID filter here.

You should also be using the ORM / filter api here not raw queries. Severe sql injection risk there. :)

code/tasks/BlogMigrationTask.php
```diff
@@ -0,0 +1,357 @@
...
+
+					$tagcount++;
+
+				}
+				// 4. Add the tag to the blogpost
+				DB::query('
+                    INSERT INTO "BlogPost_Tags"
+                    SET
+                    "BlogPostID" = \'' . $blogpostid . '\',
+                    "BlogTagID" = \'' . $tagID . '\''
```
tractorcow: Please use DB::prepared_query if you have arguments.

`DB::prepared_query("MY STRING'", array($arg1, $arg2));`

code/tasks/BlogMigrationTask.php
```diff
@@ -0,0 +1,357 @@
...
+     * Migrate ArchiveWidget
+     *
+     * We set the BlogID to the first Live BlogID that can be found as we cannot know this value from
+     * the old blog data structure.
+     *
+     * Set some default values for the NumberToDisplay as this was also not part of Blog 1.0 data.
+     */
+	protected function migrateWidgets() {
+		//Get the first Blog ID else set to 0.
+		$widgetblogid = DB::query('SELECT "ID" FROM "Blog_Live" ORDER BY "ID" LIMIT 1')->value();
```
tractorcow: This isn't going to work for multiple blogs.

code/tasks/BlogMigrationTask.php
```diff
@@ -0,0 +1,357 @@
...
+		if (!$widgetblogid) {
+			$widgetblogid = 0;
+		}
+
+		//ArchiveWidget to BlogArchiveWidget
+		//Yearly
+		DB::query('
+			INSERT INTO "BlogArchiveWidget" ("ID", "NumberToDispay", "ArchiveType", "BlogID")
+			(
+				SELECT "ID", 10, \'Yearly\', ' . $widgetblogid . '
```
tractorcow: I know that these are just numeric values, but writing SQL in this way is extremely risky. Please use the ORM if possible.

tractorcow:
Hi @camfindlay , are you still happy to continue working on this or would you like one of the open sourcerers to take on implementing my feedback?

camfindlay:
I'm happy for anyone to jump in and implement the suggested fixes (happy for people to PR against my branch and have it flow through to this PR).

dhensby:
This is now targeting the new branch `2` as `master` is SS4 compat

tractorcow:
No movement in a year, I think this PR isn't going anywhere sorry.

code/tasks/BlogMigrationTask.php
```diff
@@ -0,0 +1,357 @@
...
+		//Get the first Blog ID else set to 0.
+		$widgetblogid = DB::query('SELECT "ID" FROM "Blog_Live" ORDER BY "ID" LIMIT 1')->value();
+		if (!$widgetblogid) {
+			$widgetblogid = 0;
+		}
+
+		//ArchiveWidget to BlogArchiveWidget
+		//Yearly
+		DB::query('
+			INSERT INTO "BlogArchiveWidget" ("ID", "NumberToDispay", "ArchiveType", "BlogID")
```
Amenel: `NumberToDisplay`: there's a missing L in there.

code/tasks/BlogMigrationTask.php
```diff
@@ -0,0 +1,357 @@
...
+			(
+				SELECT "ID", 10, \'Yearly\', ' . $widgetblogid . '
+				FROM "ArchiveWidget"
+				WHERE "DisplayMode" = \'year\'
+			)
+		');
+
+		//Monthly
+		DB::query('
+			INSERT INTO "BlogArchiveWidget" ("ID", "NumberToDispay", "ArchiveType", "BlogID")
```
Amenel: Same here on `NumberToDisplay`: there's a missing L.

code/tasks/BlogMigrationTask.php
Amenel: I had to replace `.$tag.` with `.addslashes($tag).` for this to work for me. The query was failing when the tag contained a single quote.

code/tasks/BlogMigrationTask.php
```diff
@@ -0,0 +1,357 @@
...
+              (
+              	SELECT "BlogHolder_versions"."ID", "BlogHolder_versions"."RecordID", "BlogHolder_versions"."Version", 10
+              	FROM "BlogHolder_versions"
+              )
+            ');
+			echo "Migrated BlogHolder versions to Blog_versions" . $this->eol;
+		} catch (Exception $e) {
+			echo "BlogHolder to Blog migration already run, moving along..." . $this->eol;
+		}
+		//SiteTree ClassName BlogEntry to BlogPost
```
Amenel: SiteTree ClassName BlogHolder to Blog

Amenel:
@camfindlay Thanks a ton for your work on this. It has helped me finally break the 2.x barrier, as imperfect as it may seem with regards to the review standards; it was immensely useful to me.

NightJar:
@Amenel as this is a closed pull request due to lack of movement, how would you feel about picking it back up (forking the [working branch](https://github.com/camfindlay/silverstripe-blog/tree/migration-refactor) from @camfindlay) and making the finishing touches? It seems like you've found some value in it, it stands to reason you're not the only one.

code/tasks/BlogMigrationTask.php
robbieaverill: Yeah I guess `Convert::raw2sql($tag)` would be the better method to use

NightJar:
Same for you @mlewis-everley - This is closed simply through lack of movement. A fork and finish to the merge is not off the table!

mlewis-everley:
Well unfortunately I don't have a lot of time at the moment to do a PR/muck about with it, etc, but I have posted a migration task that worked for me here: https://github.com/silverstripe/silverstripe-blog/issues/440#issuecomment-454595515

NightJar:
That's how it be :)
No stress, just a gentle suggestion 👍

PR #317, opened 2015-11-04
BUG Fix crash if parent page isn't Blog type

tractorcow:
Most of the code in this class should gracefully fall back if the parent page isn't of type Blog, but this method forgot to do so.

### DIFF
--- a/code/model/BlogPost.php
+++ b/code/model/BlogPost.php
@@ -602,14 +602,26 @@ class BlogPost extends Page {
 	 * @return ArrayList
 	 */
 	protected function getDynamicCredits() {
-		$items = new ArrayList();
+		// Find best page to host user profiles
+		$parent = $this->Parent();
+		if(! ($parent instanceof Blog) ) {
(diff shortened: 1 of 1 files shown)


code/model/BlogPost.php
```diff
@@ -604,12 +604,21 @@ public function getCredits() {
 	protected function getDynamicCredits() {
```
micmania1: Do you think this return SS_List instead? If the parent isn't an instance of Blog it seems a waste to loop through all authors and create an ArrayList if you could just return the DataList.

tractorcow: Sure done.

assertchris: > Sure done.
Still seems to return an ArrayList to me. Also, this is a breaking change.

tractorcow: ArrayList is an SS_List, but I guess if user code relies on an ArrayList being returned it's now breaking. I'll revert.

PR #319, opened 2015-11-05
BUG Fix tag / category filters not being filtered in PaginatedList()

tractorcow:
It's necessary to use $blogPosts because this is the variable that contains the pre-filtered list for the current blog, which is assigned during one of the appropriate filtering methods (category / tag / etc).

### DIFF
--- a/code/model/Blog.php
+++ b/code/model/Blog.php
@@ -959,27 +959,21 @@ class Blog_Controller extends Page_Controller {
 	 * @return PaginatedList
 	 */
 	public function PaginatedList() {
-		/**
-		 * @var Blog $dataRecord
-		 */
-		$dataRecord = $this->dataRecord;
+		$allPosts = $this->blogPosts ?: new ArrayList();
+		$posts = new PaginatedList($allPosts);
 
-		$posts = new PaginatedList($this->getBlogPosts());
-
+		// Set appropriate page size
 		if($this->PostsPerPage > 0) {
-			$posts->setPageLength($this->PostsPerPage);
+			$pageSize = $this->PostsPerPage;
+		} elseif($count = $allPosts->count()) {
+			$pageSize = $count;
 		} else {
 			$pageSize = 99999;
-
-			if($count = $dataRecord->getBlogPosts()->count()) {
-				$pageSize = $count;
-			}
-
-			$posts->setPageLength($pageSize);
 		}
+		$posts->setPageLength($pageSize);
 
+		// Set current page
 		$start = $this->request->getVar($posts->getPaginationGetVar());
-
 		$posts->setPageStart($start);
 
 		return $posts;
@@ -996,7 +990,9 @@ class Blog_Controller extends Page_Controller {
 		 */
 		$dataRecord = $this->dataRecord;
 
-		$rss = new RSSFeed($dataRecord->getBlogPosts(), $this->Link(), $this->MetaTitle, $this->MetaDescription);
+		$this->blogPosts = $dataRecord->getBlogPosts();
+
+		$rss = new RSSFeed($this->blogPosts, $this->Link(), $this->MetaTitle, $this->MetaDescription);
 
 		$this->extend('updateRss', $rss);
 
--- a/tests/BlogTest.php
+++ b/tests/BlogTest.php
@@ -253,4 +253,72 @@ class BlogTest extends SapphireTest {
 		$this->assertFalse($postB->canPublish($visitor));
 		$this->assertFalse($postC->canPublish($visitor));
 	}
+
+	public function testFilteredCategories() {
+		$blog = $this->objFromFixture('Blog', 'FirstBlog');
+		$controller = new Blog_Controller($blog);
+		
+		// Root url
+		$this->requestURL($controller, 'first-post');
+		$this->assertIDsEquals(
+			$blog->AllChildren()->column('ID'),
+			$controller->PaginatedList()->column('ID')
+		);
+
+
+		// RSS
+		$this->requestURL($controller, 'first-post/rss');
+		$this->assertIDsEquals(
+			$blog->AllChildren()->column('ID'),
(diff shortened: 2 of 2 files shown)


camfindlay:
this...

micmania1:
This was introduced in #301 to fix a different bug which will be re-introduced if this is merged.

The problem is that $this->blogPosts is set by different methods depending on the request and those methods return different types. Some only return DataList and some can return null which is what caused #301.

Can you add some sort of check to ensure that $this->blogPosts isn't null before creating the paginated list?

Going forward it would be good to tie up the loose ends that caused $this->blogPosts to become null and have all methods which it depends on, return a DataList (or SS_List at least).

tractorcow:
Ok, here's a fix that resolves both issues.

tractorcow:
@micmania1 I would have not made the lists assigned in the actual method handlers; I would have put all the filtering logic in the getter, and have it inspect the current $this->request->param('Action') to determine how to generate the list. :)

Lazy-loading this list means you don't need to rely on some pre-occuring action to filter the list, and avoids breakages if this assumption is violated.

code/model/Blog.php
```diff
@@ -954,32 +954,40 @@ public function getFilterDescription() {
...
+	 *
+	 * @return SS_List
+	 */
+	public function BlogList() {
+		if($this->blogPosts) {
+			return $this->blogPosts;
+		}
+
+		// Fallback to unfiltered list
+		return $this->dataRecord->getBlogPosts();
```
micmania1: I think this changes the functionality. If there are no blog posts (where it would usually return null), it should just return an empty ArrayList otherwise blog posts will show where they're not supposed to.

camfindlay:
Would be good to get this one fixed up soon, anything I can do to help resolve @tractorcow @micmania1  ?

micmania1:
I think we just need BlogList to return an empty ArrayList() and it should be fine.

camfindlay:
Looks like this same issue might be affecting the archive filtering too.

tractorcow:
Fix incoming. :D

tractorcow:
I've changed the behaviour of the controller thusly:

- Blog_Controller::init() will now pre-initialise the `blogPosts` member variable to all posts for this blog.
- Each individual action will now reset this list to null, before potentially re-applying a filtered list for the appropriate filter.
- This means that if any actions rely on this list, there will always be an appropriate value, but if a filter fails to find any posts, it will be empty (correctly).
- Tests. :D

camfindlay:
Branch tested on large blog codebase. All working as expected. Merge I say!

tractorcow:
@micmania1 want to review this?

micmania1:
I don't like the idea of setting $this->blogPosts to the entire blog list then unsetting it in actions.

If a user extends blog and forgets to set $this->blogPosts to null then they may show the wrong data. Its important to protect against this by default incase the user has some permissions logic around what blog posts to show.

Only in the index method would you ever want to set $this->blogPosts to the entire list which is why I don't think its a reasonable default. At the most, $this->blogPosts should only be set to an empty ArrayList in the init method which is a sensible default.

Returning an empty ArrayList form BlogList when $this->blogPosts is empty/null should be enough to fix the issue.

camfindlay:
@micmania1 @tractorcow if you want to try out the suggest change here let me know happy to run against our blog site to ensure in practice things are working.

tractorcow:
> Only in the index method would you ever want to set $this->blogPosts to the entire list 

And in the `rss` method as well. As well as `atom` should we add it.

> Returning an empty ArrayList form BlogList when $this->blogPosts is empty/null should be enough to fix the issue.

This only works if we manually assign in each of the above unfiltered methods. This is why I suggested making it the default.

tractorcow:
> If a user extends blog and forgets to set $this->blogPosts to null

This is only a problem if they then forget to set it to the filtered list. Even if they do forget it, calling httpError(404) will show the error page in any case, so having the wrong list there won't matter really.

camfindlay:
sounds reasonable, perhaps merge as a solution and patch over if a better solution can be prepared? As it stands, the filtering is broken and 2.1.0 is tagged as being stable which is not such a good thing.

micmania1:
This can be fixed with a 3 line change to check that blogPosts is not null. I'm not really happy adding extra complexity into something that doesn't need it.

tractorcow:
Well, if that's the case, then let's not try to make BlogList an authoritative list. Each method can just handle their own data source.

tractorcow:
@micmania1 I've pushed up a fix that's hopefully in line with the spirit of the original code.

code/model/Blog.php
```diff
@@ -959,27 +959,22 @@ public function getFilterDescription() {
...
 	 */
 	public function PaginatedList() {
-		/**
-		 * @var Blog $dataRecord
-		 */
-		$dataRecord = $this->dataRecord;
-
-		$posts = new PaginatedList($this->getBlogPosts());
+		$allPosts = $this->blogPosts ?: new ArrayList();
+		$allPosts = $allPosts->sort(BlogPost::config()->default_sort);
```
micmania1: There's no need to sort. This can be done when the list is initially created from the request. Problem with setting it here is that it overwrites anything perviously set.

code/model/Blog.php
```diff
@@ -996,7 +991,11 @@ public function rss() {
 		 */
 		$dataRecord = $this->dataRecord;
 
-		$rss = new RSSFeed($dataRecord->getBlogPosts(), $this->Link(), $this->MetaTitle, $this->MetaDescription);
+		$this->blogPosts = $dataRecord
+			->getBlogPosts()
+			->sort(BlogPost::config()->default_sort);
+
```
micmania1: This can be the same. $posts = $this->blogPosts ?: new ArrayList();

tractorcow:
Thanks!

camfindlay:
I'll test this fresh merge out on a site ;)

PR #320, opened 2015-11-08
ENHANCEMENT Default archive year

camfindlay:
If no year is passed, rather than 404 it should grab the latest years posts.

### DIFF
--- a/code/model/Blog.php
+++ b/code/model/Blog.php
@@ -727,13 +727,17 @@ class Blog_Controller extends Page_Controller {
 	/**
 	 * Fetches the archive year from the url.
 	 *
-	 * @return null|int
+	 * @return int
 	 */
 	public function getArchiveYear() {
-		$year = $this->request->param('Year');
 
-		if(preg_match('/^[0-9]{4}$/', $year)) {
-			return (int) $year;
+		if($this->request->param('Year')){
+
+			if(preg_match('/^[0-9]{4}$/', $year = $this->request->param('Year'))) {
+				return (int) $year;
+			}
+		} else {
+			return SS_Datetime::now()->Year();
 		}
 
 		return null;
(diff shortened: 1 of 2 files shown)


code/model/Blog.php
```diff
@@ -736,7 +736,7 @@ public function getArchiveYear() {
 			return (int) $year;
 		}
 
-		return null;
+		return (int) date('Y');
```
micmania1: This should be SS_Datetime::now()->Year() so that we can unit test correctly. Can you write a test to ensure this returns the correct year? If you check the test fixtures it should tell you what the current year is.

micmania1:
I've left 1 comment. You'll need to fix the failing test so that it returns the expected 200 response now.

camfindlay:
@micmania1 working on this now

camfindlay:
@micmania1 passing and checking for a default year but will 404 if the year is invalid.

tests/BlogTest.php
```diff
@@ -109,8 +109,8 @@ public function testArchiveLinks() {
 		$this->assertEquals(200, $this->getStatusOf($link), 'HTTP Status should be 200');
 
 		$link = Controller::join_links($blog->Link('archive'));
-
-		$this->assertEquals(404, $this->getStatusOf($link), 'HTTP Status should be 404');
+		$this->assertEquals(200, $this->getStatusOf($link), 'HTTP Status should be 200');
+		$this->assertEquals(SS_Datetime::now()->Year(), ModelAsController::controller_for($blog)->getArchiveYear(), 'Defaults to current year');
```
micmania1: https://github.com/silverstripe/silverstripe-blog/pull/321 - the date is configured in setUp(). Its always 2013 :)

tractorcow: Wouldn't that just mean that SS_Datetime::now()->Year(); always returns 2013, so it's ok anyway? :P

micmania1: Yeah, but you're testing SS_Datetime::now()->Year() == SS_Datetime::now()->Year(). You might as well not test it.

PR #322, opened 2015-11-09
FIX Title bug due to #320

camfindlay:
The archive when defaulting to the current year should read the same as it someone provided the current year in a param.


micmania1:
What happens before/after this change?

tractorcow:
I think the site would think we were always filtering by year, even when not in /archive action

camfindlay:
@micmania1 @tractorcow updated and added a better test.

PR #325, opened 2015-11-11
Added featured posts to CMS

assertchris:
Added the ability to feature posts in the CMS (~~with a timestamp, for ordering purposes~~). This is part of a larger move to allow featured posts to be prominent on the front-end.


mandrew:
I'm unsure if putting the feature button anywhere near the publish buttons is a good idea? Would be good to run this by a few people to get any other ideas. Might be better to add it to the side bar? @clarkepaul have you got any other ideas?

tractorcow:
I think it's a good place for the button.

From a usability perspective, this helps a lot when writing new posts. Most of the time you want the last post you've edited to be the "featured".

micmania1:
Can this not be a module? It certainly doesn't make sense to add this as a DataExtension on BlogPost in the core module, and featured is not necessarily a must-have on all blogs.

code/extensions/BlogPostFeatureExtension.php
```diff
@@ -0,0 +1,77 @@
...
+	 *
+	 * @return FieldList
+	 */
+	public function updateCMSActions(FieldList $actions)
+	{
+		if ($this->owner instanceof BlogPost) {
+			/** @var null|SS_DateTime $featureDate */
+			$featureDate = $this->owner->dbObject('FeatureDate');
+
+			if ($featureDate->getValue()) {
```
micmania1: This will fail if $featuredDate is null. You can use `getField` instead of `dbObject` to always return `SS_Datetime`.

code/extensions/BlogPostFeatureExtension.php
```diff
@@ -0,0 +1,77 @@
...
+
+		return $this->owner->redirectBack();
+	}
+
+	/**
+	 * @param array $data
+	 *
+	 * @return bool|SS_HTTPResponse
+     */
+    public function doUnfeature($data)
```
micmania1: tabs :p

assertchris:
@micmania1 can't the extension be disabled (through config override) in sites that use the blog module? It's 1 field but a lot of change (still to come) to the front-end templates. It doesn't make sense to me to have as another module...

code/extensions/BlogPostFeatureExtension.php
tractorcow: dbObject always returns an object, even if the value is empty. I think you have it backwards?

I would suggest to use ->exists() instead of ->getValue() but the result is pretty much the same.

micmania1:
Disabling extensions is tricky because they're added in YAML as arrays. The only way to remove it is at runtime using Config->update() or Config->remove() which is REALLY bad for performance.

SpiritLevel:
Alternatively, you could put the "Feature" button in the "More Options" dropdown...the Save, Save & publish button pair then remain uniform since context sensitive actions are hidden in the dropdown. :)

tractorcow:
I think it's probably ok to have it disabled by default, but include documentation on how to enable it (it should be a simple case of adding an extension to BlogPost).

assertchris:
@tractorcow, @micmania1 good idea...

assertchris:
@SpiritLevel Interesting...Lemme think about that for a bit. :)

camfindlay:
I'm 50/50 on the button being there. Perhaps if it initially said "Save & feature" as a third option for the initial publishing action? I'd imaging that would be the largest use case. An author writing a featured post will know it's going to be featured perhaps.

camfindlay:
I don't think it should be in the 'more options' as it might be easily missed

code/extensions/BlogPostFeatureExtension.php
```diff
@@ -0,0 +1,77 @@
+<?php
+
+class BlogPostFeatureExtension extends DataExtension {
+	/**
+	 * @var array
+	 */
+	public static $allowed_actions = array (
+		'doFeature',
```
camfindlay: by whitelisting these does it means I can pings these with post data? We had an issue in forum when we had exposed handler methods.

nicolewilliams16:
From a content editor perspective I'm uncomfortable about "featured" being connected to the Save button for two reasons, this is not the normal user flow on every page (it breaks the pattern of all other pages), and secondly this is a toggle option whereas the save buttons don't work in this way.

assertchris:
@nicolewilliams16 good reasons! What about @mandrew's idea of it being in the sidebar? a button there doesn't make sense though, so probably a checkbox and some kind of timestamp mechanism...

code/extensions/BlogPostFeatureExtension.php
assertchris: I guess we should have permission checks on these...

tractorcow:
Hm, maybe it could be an inline form action, and then put it beside the "feature date" field (if such a think exists).

clarkepaul:
I don't think the timestamp is needed for the case we need, although that could be something somebody might want  - probably good to keep but I wouldn't think its the best way to create a custom order of posts anyway eg numbered, drag drop.. . If somebody goes and makes old posts featured doesn't always mean they want them to replace more recent featured posts. Blogs have a common default state of being ordered by date and I would think that is a common expectation unless the user has reordered it for their specific case.
I'm really not liking the addition of the featured button mixed with the main actions, it just doesn't feel right mixing a toggle there (I take it that's how it works). I think in the side is better personally, with either a checkbox or switch.
pasted_image_12_11_15__3_23_pm

It seems though that the label for the image either shouldn't be "Featured Image" otherwise it feels like a trigger to make a post featured. I think it needs to be renamed to something like "Main image" or just leave it as "Post image" or "Image".
pasted_image_12_11_15__3_28_pm

code/extensions/BlogPostFeatureExtension.php
micmania1: Does it not work without $allowed_actions? The action is normally triggered by the form so these don't usually need to be directly accessed.

assertchris:
I've updated this:

screen shot 2015-11-13 at 7 54 04 am

screen shot 2015-11-13 at 7 54 49 am

tractorcow:
This is good. :dancer: 

Everyone happy with this change?

nicolewilliams16:
Looks good 

Re @clarkepaul suggestion - I'd suggestion chanign the image upload heading to "Hero Image" or "Banner Image"

assertchris:
@nicolewilliams16, @clarkepaul 

screen shot 2015-11-13 at 9 30 17 am

nicolewilliams16:
perfect :)

code/model/BlogPost.php
```diff
@@ -673,6 +675,13 @@ protected function onBeforeWrite() {
 			$this->Authors()->add($member);
 		}
 	}
+
+	/**
+	 * @return bool
+     */
+	public function getCanBeFeatured() {
+		return BlogPost::has_extension('BlogPostFeatureExtension');
+	}
```
stevie-mayhew: Unused function?

stevie-mayhew:
It would be good to have `getFeaturedPosts` and then use `$FeaturedPosts.First` in the template, rather than making it only available as a single post. Its a bit more flexible.

code/model/BlogPost.php
stevie-mayhew: I also feel like this should be in an extension as its not part of the "core" code.

assertchris:
@stevie-mayhew, @tractorcow, @camfindlay 

Added the ability to specify how many featured posts should be removed from normal lists, in Blog → Settings

screen shot 2015-11-13 at 10 58 27 am

code/extensions/BlogFeatureExtension.php
```diff
@@ -0,0 +1,72 @@
...
+	 */
+	private static $defaults = array(
+		'ExcludedFeaturePosts' => 1,
+	);
+
+	/**
+	 * @return bool
+     */
+	public function getCanHaveFeaturedBlogPosts()
+	{
```
micmania1: SilverStripe coding conventions (brackets on the same line)? :p

assertchris: https://groups.google.com/forum/#!topic/silverstripe-dev/_ebPef7HgKQ

screen shot 2015-11-13 at 11 38 12 am

camfindlay: Use case for this? Seems like something that might be better as a yaml config? Or perhaps it sets itself based on the number of featured posts asked for to display in the feature posts hero banner (or whatever the project calls for).

stevie-mayhew:
You shouldn't sneak code style changes into an unrelated PR.

code/extensions/BlogFeatureExtension.php
stevie-mayhew: I think it should be "Featured Posts To Display" and an option then to "Remove Featured Posts From Normal List" rather than this exclusion which is confusing.

assertchris: I have moved this to a config setting.

tractorcow:
I appreciate the hard work @assertchris is doing to respond to all of this feedback. :)

assertchris:
@mandrew, @camfindlay @stevie-mayhew, @micmania1, @tractorcow: any more feedback, or is it time to merge?

camfindlay:
other than 1 knitpicky comment from me, I think this is looking good

assertchris:
@camfindlay I don't think we need to rename FeatureImage to BannerImage. By changing the title, we satisfy @clarkepaul's concern without introducing a BC break. If we're afraid to change field titles independently of their names then I don't see the point in having modifiable titles...

camfindlay:
yield, I'm not going to push on that one so I'm happy with the feature, feature :)

code/extensions/BlogFeatureExtension.php
micmania1: I'm all for PSR-2, but the rest of this module isn't PSR-2. I'd be happy to adopt it, but only if it was adopted everywhere.

SpiritLevel:
Looks good! Putting "featured" Inline make more sense than putting it in the More Options dropdown. 

A small idea from the peanut gallery: The phrase "Mark as featured post" with a checkbox feels a bit clunky. Perhaps just "Feature" or "Feature Post" followed by radio buttons labelled "Yes" and "No", and defaulting to "No" ? Or a switch ? Just a thought :)

assertchris:
@SpiritLevel interesting, but perhaps as a different pull request. This one has had enough traffic for now, I think :)

stevie-mayhew:
Still concerned with the function `getCanBeFeatured` which isn't used.

docs/en/featured-posts.md
```diff
@@ -0,0 +1,38 @@
...
+![](_images/featured-posts-cms.png)
+
+By default, the template will show the most recent featured post at the top of the
+list of posts in a blog. This post will be removed from the normal list of blog posts.
+You can increase the number of specially-displayed feature posts by modifying the
+template to show more, and by changing the following config setting:
+
+```
+<% if $CanHaveFeaturedBlogPosts && $FeaturedBlogPosts %>
+	<% loop $FeaturedBlogPosts %>
```
stevie-mayhew: This should have a limit set of it which matches the `excluded_feature_posts` in some way.

If you have a blog displaying 10 posts and you have featured 10/10 of the last posts, and only decide to exclude 3, then you will have 7 extras showing here which are duplicates of the ones below.

assertchris:
@stevie-mayhew it's a definitive answer to "is IsFeatured returning false because the extension is disabled or because this post isn't featured?". Just because none of the templates is using it, does not mean it is useless.

stevie-mayhew:
@assertchris you're right, its not useless. However it isn't necessary either.

assertchris:
@stevie-mayhew I think I want to keep it.

micmania1:
I'm with @stevie-mayhew on the getCanBeFeatured methods. These can be moved to Blog/BlogPost to create a consistent API, whether the extension is enabled or not.

For example, if the extension is not applied  then calling $blog->getCanBeFeatured() will error. If you move it to Blog and add the logical check in there, then people can use the API without worrying about whether the method is available.

assertchris:
@micmania1 actually, I think @stevie-mayhew's point was to move those out of the main code and into an extension. Those methods are created as setters in templates. If the extensions are not enabled then the values of those methods will be false inside a template.

micmania1:
oh, then I disagree with @stevie-mayhew. 

But I do think they should belong in the main classes for the mentioned reasons. Even if the intended use-case is templates, its still a PHP API first.

assertchris:
@micmania1 featured posts are an optional addition (though one could argue they were erroneously omitted). I agree with @stevie-mayhew that keeping the opt-in feature self-contained is a good idea. And regardless, I already made the change hours ago. I can't keep painting this bike shed...

stevie-mayhew:
I was actually talking about removing `getCanBeFeatured ` completely :smile: It makes more sense for them to be in the main code but then again I don't think it makes sense anywhere :wink: 

Keeping the opt-in feature self contained is a good idea, which makes the `getCanBeFeatured` argument from your end moot - if you don't opt in then you wouldn't call it.

assertchris:
@stevie-mayhew, @micmania1 I have removed the methods, and adjusted the template. Do you have any meaningful changes you would still like me to make, or can we consider this horse flogged?

micmania1:
Can you explain why a DataExtension was used in the first place? We have Hierarchy and Versioned in the framework which supply required API for SiteTree to work, which is ridiculous. I don't want to end up in the same situation.

The entire feature can be added to Blog and BlogPost. `getFeaturedBlogPosts()` can be its own method on Blog removing the need for the new `$context` parameter on `getBlogPosts()`. It can return blog posts depending on the return value of `getCanHaveFeaturedBlogPosts()`. The CMS fields can also be added in the same way. 

Using a DataExtension just doesn't make sense to me at all. Self containment is not what's happening here - these things are being applied to a single class, but in two different files, only because the framework allows it.

Having methods which may or may not exist does not seem like a good API decision.

assertchris:
@micmania1 I am on the verge of closing this pull request and using a personal fork for what I am trying to achieve. Others have reviewed this architectural approach and do not have the same grave concerns as you do about how I have implemented it. Does anybody else think I should redo this pull request? @tractorcow @stevie-mayhew @camfindlay @mandrew

clarkepaul:
I'm happy with the direction this is going, thanks for moving the trigger @assertchris  - would also think shortening the text is a nice to have but can be done separately.

micmania1:
@tractorcow you seem to have accidentally pressed the merge button. Can you revert until we've finished our discussion?

tractorcow:
Hi @micmania1 sorry for closing when you had unresolved concerns. I personally am pretty happy with the feature as it was implemented; I should have noticed your outstanding issues though.

I think the fact that the feature is in a separate extension isn't necessarily a problem, because as soon as it's included in the core class, it no longer becomes an optional feature and is now mandatory.

How about instead we open further discussion with a pull request on top of this? If the concern is just how the functionality is implemented (rather than how the feature actually works) then I'd be happy if you wanted to refactor this. :)

PR #326, opened 2015-11-13
Removed sneaky merge featured posts extensions.

micmania1:
#325 - replacement of sneaky merge.

What this includes:
* Allows featured posts to be switched on and off per blog via the CMS - not globally via config
* Featured posts turned off by default
* No overhead by using extensions
* No sneaky merges to benefit internal projects only
* Welcoming of constructive feedback and criticism

@tractorcow @assertchris @stevie-mayhew


assertchris:
Grow up.

tractorcow:
Hi @micmania1 appreciate your taking the time to add your suggestion.

I think the primary concern here is: keep this functionality in an extension, or accept this and merge into the chore classes.

I'm happy to step out this time and let those who wish to have their say vote either way, with civil discussion of course.

At the end of the day, not everyone will have their way, but everyone will get the chance to have their voice heard. I apologise for my lack of listening earlier. :)

jonom:
Personally I'd prefer Featured Posts to be an an extension. I'm not sure this feature will see majority use, and implementing this functionality in an extension means developers are free to ignore it if they want and create a similar solution without risking a conflict with baked-in functionality.

For instance I'm not personally likely to use this approach as I don't really like tying a feature status to an individual item - I'd rather place this in context of where the posts are being featured and recognise that you might feature different posts in different areas. So I might have a $many_many relationship between HomePage and BlogPost that lets you pick a maximum of 4 or so posts to feature there.

When you have a Feature flag directly on BlogPost it's easy to end up with the majority of posts being featured because you never go back to un-feature them, as editing the list of featured posts is a bit cumbersome process.

camfindlay:
Hey looks like there is some disagreement on the implementation here.

I can see use cases for both approaches here, is there some common ground perhaps that can be found to bring together both ideas?

I'd also take this time to remind all of the [community code of conduct](https://docs.silverstripe.org/en/contributing/code_of_conduct/), please give it a look over before continuing the debate here about this feature.

tractorcow:
In any case, this commit needed to be renamed before out can be merged. As it stands it's not an appropriate revision description.

camfindlay:
@micmania1 happy for the talk on this to continue and I'll have to ask you amend revision description as @tractorcow mentions so we can keep to the implementation issues under discussion other wise might have to consider closing. So far have only had @jonom chime in for a +1 extension approach.

jonom:
I know this module is used a lot by the community and originated there so maybe it's a special case - but in general I thought add-ons weren't subject to the same merge policies as CMS and Framework? Do we need some documentation on this (or have I just missed it)?

tractorcow:
@micmania1 you seem to have a funny idea in your head that I some how colluded with @assertchris to sneak a feature into core for an internal project, something I'm not even involved in.

In addition, we've learned well enough from that mistake that was made more than a year ago. We've since introduced semantic versioning, and it's been a success both in core and in our supported modules. I see no reason why we've failed in this case to adhere to it, either. The next version of blog will be 2.2 to represent the new features we've added.

I'm sorry that you disagree with our approach, or with @assertchris 's feature, but at the end of the day, not everyone is going to get their way. There's no point getting upset or throwing out accusations because you didn't.

micmania1:
@tractorcow I never said anything about Chris colluding with you. Its clear you merged the PR with total disregard to my comments - you admit this yourself in the PR. There is no conspiracy so I think you should take your own advice re: accusations.

In terms of the technical solution - this has gone way beyond that. There are features in the blog which I never wanted there, but have since changed my mind on. I'm more than happy to do that, but in this case no effort was taken to try and explain to me why this was needed in core, even though it was the first thing I asked.

Its clear that there are others who are happy with the solution. If that's the case then I would have definitely not complained about the merge. The reason I'm not happy is because I wasn't even able to have a discussion about it.

tractorcow:
`No sneaky merges to benefit internal projects only` is the "colluding false accusation" I'm talking about, regardless of what words you feel best describe it.

I missed your comment and it's been apologised for. Your response has been disappointing to say the least, @micmania1.

micmania1:
Ignored me and merged + no revert = sneaky.

The disappointment seems to be mutual. Its not the first time from my point of view either and i'm sure it won't be the last.

camfindlay:
We are no longer on topic of the original pull request (which has since been closed), and GitHub is less of a place to air a grievance. I'd like to again remind all here of the [code of conduct](https://docs.silverstripe.org/en/3.2/contributing/code_of_conduct/) and ask that you either discuss this in-person to work things out or you are welcome to raise with me and I am happy to help mediate any dispute. I'd ask for all to refrain from using this closed pull request for off-topic discussion.

camspiers:
What is being discussed is the process that was used to include the original change. That process is relevant to the community, and people in the community do care and have opinions about it.

@camfindlay discussing it in private doesn't help other interested parties participate.

camfindlay:
@camspiers I agree with you and this was getting off the topic of process (as you rightly raise) and entering personal realms.

stevie-mayhew:
I've held off commenting on this as I feel pretty strongly about the whole thing and am afraid that will pervade my response. What I will say is that I'm disappointed that there does appear to be miscommunication of intent from most parties involved stemming from the "quick merge" after a *lot* of discussion was held in the original commit. 

I feel like I didn't get a chance to respond even though there was a call for this [shortly before the merge](https://github.com/silverstripe/silverstripe-blog/pull/325#issuecomment-156298011).

This is a concerning pattern that I've personally highlighted before to other developers inside and outside of SilverStripe where small and large changes are pushed through in what seems to be a need within a SilverStripe internal project. I'm not saying that its the case for this one, but it sure seems like it. Its also not a bad thing as it progresses SilverStripe the open source project for the community.

However, as owners of the open source project with needs for commercial projects, these 'feature PRs' should be carefully managed. In this case it clearly hasn't been carefully managed.

Architecturally there are concerns for me as well. I don't see the benefit of having this in an extension, also [raised here](https://github.com/silverstripe/silverstripe-blog/pull/325#issuecomment-155954015), which was never directly resolved. This is clearly user land code and while its useful for the blog I don't see a clearly defined majority of users needing it. It could have easily gone into a module and have been suggested via composer installation and in the documentation of the main blog module. Having an extension available which doesn't do anything by default is of course a good idea if you don't want to use it, but by definition I think it makes its inclusion redundant and furthers the argument that it should be modularised.

dhensby:
> DataExtensions are great for adding database fields to DataObjects...

@helpfulrobot  no, YML is better for that.

dhensby:
I agree with @micmania1 and @stevie-mayhew - I'm usually not too tactful with this kind of thing, but I'll try to be.

An extension that is not in use by default is a bad idea; it will be underused and under-tested. It requires maintenance for a feature we admit isn't needed regularly and the responsibility to test and maintain it becomes diluted. I also agree with @micmania1 that neither solution is satisfactory.

This looks like something that would be perfectly acceptable as an add-on module for blog.

@camfindlay and others shouldn't mistake @micmania1's frustration as counter-productive; it's a sign of his passion about SilverStripe and maintaining credibility for SilverStripe against perceptions that we merge in internal requirements with little regard for others.

To have it merged in after a day was a mistake - especially when a call for opinions was made; @tractorcow has admitted this, but the correct response is then to revert, re-open and wait for feedback.

I was quite close to unilataraly reverting the PR because it was not the right process; the time from @assertchris' call for feedback to merge was 24mins - clearly a far too short period of time.

I forward the motion that PR #325 is reverted.

I separately forward the motion that the feature is made it's own module.

PR #327, opened 2015-11-17
Added hook for extension filters on category blog posts

assertchris:


tractorcow:
Also see https://github.com/silverstripe/silverstripe-blog/pull/327

code/model/BlogCategory.php
```diff
@@ -40,6 +40,17 @@ class BlogCategory extends DataObject implements CategorisationObject {
 	private static $extensions = array(
 		'URLSegmentExtension',
 	);
+	
+	/**
+	 * @return DataList
+	 */
+	public function BlogPosts() {
+		$blogPosts = parent::BlogPosts();
```
camfindlay: this this not be simply $this->BlogPosts()? as there is a named relation here

assertchris: I think calling `$this->method()` (every time) inside `function method` causes an infinite loop.

assertchris: `parent::BlogPost()` does exist through `__call` on DataObject. It's pretty standard to override methods on child classes, and still call inherited functionality on parent classes. In this case, the inherited functionality is the code that generates the relationship DataList, and the overridden functionality is that which provides the hook.

tractorcow: Yes I've used this before, and can confirm it falls back to `__call`

PR #328, opened 2015-11-17
Added hook for extension filters on tag blog posts

assertchris:


tractorcow:
Also see https://github.com/silverstripe/silverstripe-blog/pull/327

PR #330, opened 2015-11-18
Add changelog for 2.2.0

tractorcow:
Please tag 2.2.0 if this is merged.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,7 +17,7 @@
 	},
 	"extra": {
 		"branch-alias": {
-			"dev-master": "2.0.x-dev"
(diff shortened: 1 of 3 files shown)


composer.json
```diff
@@ -17,7 +17,7 @@
 	},
 	"extra": {
 		"branch-alias": {
-			"dev-master": "2.0.x-dev"
+			"dev-master": "2.x-dev"
```
tractorcow: This should be easier than updating the minor alias every release. ;D

tractorcow:
Hold off on tagging 2.2.0 for a bit.

PR #332, opened 2015-11-18
Added standard Travis config

helpfulrobot:


dhensby:
the error _looks_ unrelated, but only happens on this PR...

micmania1:
Maybe there's an issue upstream with postgres and using the Group keyword. I don't really know enough about it.

It does highlight an issue with the actual code though. It should check for the existence of Group before calling ->Members().

robbieaverill:
This should be closed now, those branches aren't relevant any more.

PR #338, opened 2015-11-20
french translation

lekoala:
not perfect, but better than nothing :-)

### DIFF
--- a/lang/fr.yml
+++ b/lang/fr.yml
@@ -1,0 +1,117 @@
+fr:
+  ArchiveWidget:
+    PLURALNAME: 'Widget archive'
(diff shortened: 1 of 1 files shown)


dhensby:
@tractorcow how does this work with transifex?

tractorcow:
We can merge them manually, but we really prefer that people add their translations at https://www.transifex.com/silverstripe/silverstripe-blog/

tractorcow:
I've synced this with transifex; See https://github.com/silverstripe/silverstripe-blog/commit/a4c6e5abb6cacbec0a94ccceff17886185b84814

PR #343, opened 2015-11-30
This leaves a empty p tag on the page

MichaelJJames:
This leaves a empty p tag on the page

### DIFF
--- a/templates/Includes/PostSummary.ss
+++ b/templates/Includes/PostSummary.ss
@@ -13,10 +13,11 @@
 	</p>
 
 	<% if $Summary %>
-		<p>$Summary
+		$Summary
 	<% else %>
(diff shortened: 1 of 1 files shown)


micmania1:
@MichaelJJames thanks for submitting this and sorry it hasn't been looked at yet.

With this fix the styling may look different depending on whether Excerpt or Summary is used. The core problem is that Summary returns HTMLText (wrapped in p tags) and Excerpt returns raw text (no p tags). 

Would you be able to update/check your change to ensure the p tags are used consistently (ie. wrap Excerpt in p tags if it already isn't)?

micmania1:
I tested this and the p tags were missing for excerpt. Raised a new PR based of this (#401) so closing.

PR #349, opened 2015-12-08
DOCS Rework documentation to have clear user guide and dev docs.

camfindlay:
As per the supported module standard we should be making user guide related docs clear. This PR moves documentation relevant only to CMS users into the correct format. I have also added some new docs to make them more complete (I'm sure more could be added too).

Note I have included optional short user docs about comments and widgets here as the modules that provide them do not have user guides at all. I think it useful to maintain these here under this module rather than further split the user experience for reading the docs. We are also looking into better including  these supported modules in the userhelp website. This new format goes a way towards making that possible.


camfindlay:
thanks for the peer. Will tidy up tomorrow.

camfindlay:
@spiritlevel try this now I will leave the backtick code blocks for now this is acceptable GFMD format.

camfindlay:
hold off merging, just working out a convention as to whether we use "user-guide" or "userguide" as the folder name.

camfindlay:
@tractorcow this can go in like this. Can deal with any naming conventions later.

docs/en/index.md
```diff
@@ -0,0 +1,11 @@
+# Blog documentation
+
+## Development
+ * [Configuring blog posts in Sitetree](configuring-blog-posts.md)
```
tractorcow: Does docsviewer work with .md extensions now?: P

tractorcow:
I actually like that you've dumped the "_images is in root" convention. I don't think it really worked in practice with deeply nested docs.

docs/en/configuring-blog-posts.md
```diff
@@ -0,0 +1,16 @@
+## Configuring blog posts in SiteTree
+
+Because your blog is part of the SiteTree, usage is the same as any other page.
+
+By default, blog posts don't appear in the SiteTree, to avoid clutter. Instead they appear inside your blog as a GridField.
+
+![](docs/en/_images/blog-post-management.png)
```
tractorcow: Links don't work with docs/en in them

docs/en/index.md
camfindlay: yes. this also allows the docs to function in the github repo.

docs/en/configuring-blog-posts.md
camfindlay: thanks, I'll fix this.

camfindlay:
How's that @tractorcow ?

tractorcow:
Good with me. Someone else can review and merge though, in case someone begs to differ. ;)

tractorcow:
@assertchris or @micmania1 want to approve this?

assertchris:
See https://groups.google.com/forum/#!topic/silverstripe-dev/GrSOZZPgEIk

camfindlay:
It's ~ 50/50 split on the naming I think we should allow either (not to mention other modules are now using "userguide" directory now anyway.

camfindlay:
@tractorcow this should be good to merge

PR #354, opened 2015-12-21
Fix bad call to getminInnerWidth()

nimeso:
This is not a function :(

### DIFF
--- a/js/cms.js
+++ b/js/cms.js
@@ -94,7 +94,7 @@
 				this.updateLayout();
 				
 				// If this panel is open and the left hand column is smaller than the minimum, contract it instead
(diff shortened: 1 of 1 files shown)


tractorcow:
Actually, it is. If you check the entwine documentation http://hafriedlander.github.io/jquery.entwine/tutorial/step_05.html this is the standard way to access these properties.

The correct fix for the known issue is to adjust the capitalisation of the property, as it's meant to have a leading capital letter.

nimeso:
@tractorcow thanks!

dhensby:
So... Closing?

tractorcow:
Right, thanks @dhensby :D

Issue #210, opened 2015-04-29
Success message overflows


micmania1:
![image](https://cloud.githubusercontent.com/assets/881537/7388348/6e4371c2-eeb8-11e4-9de1-bc9876b9faa8.png)

Runs behind the right menu.

micmania1:
Also displays space under the right menu when you scroll, but only when the success message above is present.

![image](https://cloud.githubusercontent.com/assets/881537/7388364/9a826086-eeb8-11e4-9558-8d714770aab6.png)

assertchris:
Beta is a time for growth-hacking, not fixing things...

micmania1:
FYI this isn't possible with CSS. The HTML kind of looks like this.

```
<div class="cms-thing">
    <p class="message">Page created blah</p>
    <fieldset>
        <div class="blog-sidebar">...</div>
        <div class="blog-fields">...</div>
    </fieldset>
</div>
```

The only part of the template we control is inside fieldset and unfortunately CSS doesn't have parent selectors.

PR #358, opened 2016-01-19
Minor: Stopped messages overflowing in blog-post

lry6003:
added line to cms.js that adds margin to messages on the new blog post page in CMS

https://github.com/silverstripe/silverstripe-blog/issues/210

6e4371c2-eeb8-11e4-9de1-bc9876b9faa8

screenshot from 2016-01-20 11 29 06

### DIFF
--- a/js/cms.js
+++ b/js/cms.js
@@ -91,7 +91,7 @@
 			minInnerWidth: 620,
 			onadd: function() {
 				// Adding margin to prevent post options from overlaping message box
-				$('#Form_EditForm_error ').attr('style', 'margin-right: 56px');
+				$('#Form_EditForm_error ').css('margin-right', '56px');
 
 				this._super();
 				this.updateLayout();
--- a/js/cms.js
+++ b/js/cms.js
@@ -91,7 +91,7 @@
 			minInnerWidth: 620,
 			onadd: function() {
 				// Adding margin to prevent post options from overlaping message box
-				$('.message').attr('style', 'margin-right: 56px');
+				$('#Form_EditForm_error ').attr('style', 'margin-right: 56px');
 
 				this._super();
 				this.updateLayout();
--- a/js/cms.js
+++ b/js/cms.js
@@ -44,7 +44,7 @@
 				$this.parent().next('.description').addClass('toggle-description-correct-description');
 			}
 		});
-		
+
 		/**
 		 * Custom merge actions for tags and categories
 		 */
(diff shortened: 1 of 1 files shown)


tractorcow:
Just a question; What would have made this impossible to do with css? I guess you tried it and it didn't work for some reason?

js/cms.js
```diff
@@ -81,18 +81,21 @@
...
 		 * Customise the cms-panel behaviour for blog sidebar
-		 * 
+		 *
 		 * see LeftAndMain.Panel.js for base behaviour
 		 */
 		$('.blog-admin-sidebar.cms-panel').entwine({
 			minInnerWidth: 620,
 			onadd: function() {
+				// Adding margin to prevent post options from overlaping message box
+				$('.message').attr('style', 'margin-right: 56px');
```
tractorcow: This will affect all messages, not just the alert box. E.g. per-field validation messages also match this.

lry6003:
The problem was caused by the post options menu on the right of the page. This menu isn't on all the pages that the success message is the CSS could not be modified without changing it on all pages

js/cms.js
```diff
@@ -81,18 +81,21 @@
...
 		 * Customise the cms-panel behaviour for blog sidebar
-		 * 
+		 *
 		 * see LeftAndMain.Panel.js for base behaviour
 		 */
 		$('.blog-admin-sidebar.cms-panel').entwine({
 			minInnerWidth: 620,
 			onadd: function() {
+				// Adding margin to prevent post options from overlaping message box
+				$('#Form_EditForm_error ').attr('style', 'margin-right: 56px');
```
tractorcow: Ok one more thing, please use .css('prop', 'val') to set css, as this would otherwise override any existing style property (e.g. style="display: none").

tractorcow:
Your explanation seems reasonable to me, sorry to second guess you.

If you can make one more change I'll merge. Thanks again!

dhensby:
I'm not a fan of this, we have a magic number of 56px which could be different if someone did customise the CSS or later on if someone does update the CSS in another PR...

We should have some way of assigning a class somewhere that indicates the side bar is there and then adjust it in the stylesheet.

tractorcow:
I think @dhensby 's solution makes sense... sorry but can you follow up on this, @lry6003 ?

dhensby:
Is it possible to add a class to the page/fieldset/form/some-parent-element that dictates there is a side bar and thus assigns the margin

```scss
.has-sidebar-filter {
    & .notice {
        margin-right: 56px;
    }
}
```

Or whatever, that kind of idea

micmania1:
FYI this isn't possible with CSS. The HTML kind of looks like this.

```
<div class="cms-thing">
    <p class="message">Page created blah</p>
    <fieldset>
        <div class="blog-sidebar">...</div>
        <div class="blog-fields">...</div>
    </fieldset>
</div>
```

The only part of the template we control is inside fieldset and unfortunately CSS doesn't have parent selectors.

The fix here won't work either because the width of the success message is variable depending on whether the side panel is open or closed.

micmania1:
@lry6003 thanks for your fix and contribution, however this has been fixed slightly differently in #370.

Issue #361, opened 2016-01-21
Travis Build is Broken for Postgres


gordonbanderson:
I made a branch with minor changes to check the blog module against Postgres and Silverstripe 3.2 - it appears Postgres and 3.1 is broken, fixtures are not loading (I have not tested Postgres and 3.2 yet)

Changes:
https://github.com/gordonbanderson/silverstripe-blog/blob/postgrestest/.travis.yml

For build report see https://travis-ci.org/gordonbanderson/silverstripe-blog/builds/103890142 - only Postgres fails.  The errors appear related to an issue with the fixtures.  See https://travis-ci.org/gordonbanderson/silverstripe-blog/jobs/103890143 for log

tractorcow:
```
pg_query(): Query failed: ERROR:  invalid input syntax for integer: ""
LINE 1: ...g"."URLSegment" = 'first-tag') AND ("BlogTag"."BlogID" = '')
```

This error is real; Where are you getting this? It could be a bug in blog that's breaking this.

tractorcow:
Ah, here's the problem.

https://github.com/silverstripe/silverstripe-blog/blob/master/code/extensions/URLSegmentExtension.php#L45

It should check to see if a BlogID is assigned, and not simply assume it is.

tractorcow:
Postgres is good... it lets us know when we make mistakes and tries to help us. It's only being a buddy.

gordonbanderson:
The problem may be in the fixtures generation and not the blog module.  I've changed the start of the method you highlighted to this:

```
public function generateURLSegment($increment = null)
    {
        $filter = new URLSegmentFilter();

        $this->owner->URLSegment = $filter->filter($this->owner->Title);

        if (is_int($increment)) {
            $this->owner->URLSegment .= '-' . $increment;
        }

        error_log('Creating URLSegment for ' . $this->owner->Title);
        error_log("\t{$this->owner->ClassName}");
        error_log("\t{$this->owner->BlogID}");

        if (!$this->owner->BlogID) {
            error_log("\t++++ BLOG ID OF 0 ++++");
            asdfsdfsdf;
        }
```

and also added trace in FixtureFactory.php, trace that confirms an ID exists for the Blog prior to the error.  With the above failing on `asdfsdfsdf` deliberately, stopping the single test from running, I can that in the MySQL database does indeed have four entries for Blog in both SiteTree and Blog tables.

[...]

PR #362, opened 2016-01-22
FIX: Tests now pass in Postgres on 3.1 and 3.2

gordonbanderson:
Problems resolved:
1) Case sensitivy of check on Code field of Group
2) MySQL and Postgres have different date functions
3) When BlogID is empty, explicitly set it to 0.  If not then all tests break under Postgres

Fix for https://github.com/silverstripe/silverstripe-blog/issues/361

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,13 +28,18 @@ matrix:
       env: DB=MYSQL COVERAGE=1
     - php: 5.5
       env: DB=MYSQL
+    - php: 5.6
+      env: DB=PGSQL
+    - php: 5.6
+      env: DB=MYSQL CORE_RELEASE=3.2
+    - php: 5.6
+      env: DB=PGSQL CORE_RELEASE=3.2
     - php: 5.4
       env: DB=MYSQL
     - php: 5.3
       env: DB=MYSQL
     - php: hhvm
       env: DB=MYSQL
-      before_install:
 
 
 before_script:
--- a/code/extensions/URLSegmentExtension.php
+++ b/code/extensions/URLSegmentExtension.php
@@ -20,7 +20,9 @@ class URLSegmentExtension extends DataExtension
      */
     public function onBeforeWrite()
     {
-        $this->owner->generateURLSegment();
+        if ($this->owner->BlogID) {
+            $this->owner->generateURLSegment();
+        }
     }
 
     /**
@@ -38,6 +40,12 @@ class URLSegmentExtension extends DataExtension
 
         if (is_int($increment)) {
             $this->owner->URLSegment .= '-' . $increment;
+        }
+
+        // Postgres use '' instead of 0 as an emtpy blog ID
+        // Without this all the tests fail
+        if (!$this->owner->BlogID) {
+            $this->owner->BlogID = 0;
         }
 
         $duplicate = DataList::create($this->owner->ClassName)->filter(array(
--- a/code/model/Blog.php
+++ b/code/model/Blog.php
@@ -473,15 +473,37 @@ class Blog extends Page implements PermissionProvider
 
         $query->innerJoin('BlogPost', sprintf('"SiteTree%s"."ID" = "BlogPost%s"."ID"', $stage, $stage));
 
-        $query->where(sprintf('YEAR("PublishDate") = \'%s\'', Convert::raw2sql($year)));
+        // getConn is deprecated, but not get_conn in 3.1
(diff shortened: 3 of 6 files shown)


dhensby:
this needs rebasing :(

if you run `git pull --rebase https://github.com/silverstripe/silverstripe-blog master` then `git push -f` it should work

gordonbanderson:
Rebase done as requested

dhensby:
thanks :)

.travis.yml
```diff
@@ -43,6 +48,8 @@ before_script:
   - git clone git://github.com/silverstripe-labs/silverstripe-travis-support.git ~/travis-support
   - php ~/travis-support/travis_setup.php --source `pwd` --target ~/builds/ss
   - cd ~/builds/ss
+  - cat ./framework/model/DB.php
```
dhensby: do we need these lines?

gordonbanderson: No, that is debug I thought I'd rebased out.  Fix, rebase and git push -f again...

gordonbanderson:
Late my time so if any further issues will have to wait until Monday realistically.

code/extensions/URLSegmentExtension.php
```diff
@@ -40,6 +42,12 @@ public function generateURLSegment($increment = null)
...
 
+        // Postgres use '' instead of 0 as an emtpy blog ID
+        // Without this all the tests fail
+        if (!$this->owner->BlogID) {
+            $this->owner->BlogID = 0;
+        }
+
         $duplicate = DataList::create($this->owner->ClassName)->filter(array(
             'URLSegment' => $this->owner->URLSegment,
             'BlogID' => $this->owner->BlogID,
```
tractorcow: If we know BlogID is 0 why are we filtering on it? It won't match anything. The second condition should have been inside an `if()` similar to how the `$this->owner->ID` check is.

In any case, the code path won't really reach here if BlogID isn't set, so it shouldn't really matter in any case.

code/model/Blog.php
```diff
@@ -473,16 +473,38 @@ public function getArchivedBlogPosts($year, $month = null, $day = null)
...
+            if ($month) {
+                $where .= sprintf(' AND EXTRACT(MONTH FROM "PublishDate") = \'%s\'', Convert::raw2sql($month));
+
+                if ($day) {
+                    $where .= sprintf(' AND EXTRACT(DAY FROM "PublishDate") = \'%s\'', Convert::raw2sql($day));
+                }
+            }
+
+            $query->where($where);
         }
```
tractorcow: Else, it just doesn't work? So now blog only works on postgres and mysql?

@dhensby I'm disappointed in this getting merged. =/

@gordonbanderson this is a good try, I'm not blaming you. This should have been given some more review prior to merge, that's all.

tractorcow:
@gordonbanderson thanks for your hard work to try to diagnose this issue. I'll suggest a change to this fix later on today.

tractorcow:
`DB::getConn()->formattedDatetimeClause` is the best way to delegate date formatting syntax to the current connector. Maybe we can use that instead?

tractorcow:
I've thrown up a possible fix at https://github.com/silverstripe/silverstripe-blog/pull/363; I'm just going to let travis-ci take care of all the cross-adaptor testing for me though. :)

code/extensions/URLSegmentExtension.php
gordonbanderson: Take those 3 lines out and all of the Postgres tests fail.  The issue does not occur with MySQL and SQLite.

code/model/Blog.php
gordonbanderson: Previously it only worked with MySQL... re-testing I forked prior to my recent changes and added Travis checks (which were missing) for PGSQL and SQLite, see https://github.com/gordonbanderson/silverstripe-blog/commits/master_20160119_check_dbs

The Travis build fails for both Postgres and SQLite, https://travis-ci.org/gordonbanderson/silverstripe-blog/builds/104570379

gordonbanderson:
I was previously unaware of the `DB::getConn()->formattedDatetimeClause` method

dhensby:
derp :/ sorry - I was obviously a bit too keen when merging this and thought green ticks meant all the DBs were ok. That if/else was quite obvious now looking back.

note to self: must try harder

tractorcow:
Sorry, I didn't mean to come off sounding harsh. :) The process works... peer review saved the day. :P

PR #363, opened 2016-01-24
BUG Fix blog date filtering for additional database server types

tractorcow:
I've also locked this to the 3.x major version to prevent semver regressions in master.

To that extend, it's safe to use DB::getConn(), as it's not going to be removed until 4.0.0.


tractorcow:
@dhensby is this fix ok in substitution for the one at #362?

dhensby:
much better

Issue #364, opened 2016-01-25
Subject Header for Blog Post Notification Missing BlogPost Title


gordonbanderson:
The offending line is https://github.com/silverstripe/silverstripe-blog/blob/master/code/extensions/BlogPostNotifications.php#L30 - it's missing a %s to substitute the BlogPost title into

gordonbanderson:
Fix here https://github.com/gordonbanderson/silverstripe-blog/commit/687406dea8770bb9d00df856d00c317f1d6a273d but I'll let Travis run it's thing first prior to issuing pull request.

micmania1:
Thanks for the fix :+1:

PR #365, opened 2016-01-25
FIX: Missing Blog Post Title from Comment Notification

gordonbanderson:
Fix for https://github.com/silverstripe/silverstripe-blog/issues/364


tractorcow:
Thanks @gordonbanderson !

PR #367, opened 2016-01-26
FIX: Add comment notifications module to composer suggested packages

gordonbanderson:
The above package is used for blog post notifications

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -29,7 +29,8 @@
 	],
 	"suggest": {
 		"silverstripe/widgets": "Some widgets come with the blog which are compatible with the widgets module.",
-		"silverstripe/comments": "This module adds comments to your blog."
+		"silverstripe/comments": "This module adds comments to your blog.",
+        "silverstripe/comment-notifications": "Get email notifications of comments on blog posts"
 	},
 	"replace": {
 		"micmania1/silverstripe-blog": "*"


composer.json
```diff
@@ -29,7 +29,8 @@
 	],
 	"suggest": {
 		"silverstripe/widgets": "Some widgets come with the blog which are compatible with the widgets module.",
-		"silverstripe/comments": "This module adds comments to your blog."
+		"silverstripe/comments": "This module adds comments to your blog.",
+        "silverstripe/comment-notifications": "Get email notifications of comments on blog posts"
```
dhensby: sorry to be so pedantic, but please can you use tabs for this like the other lines?

gordonbanderson:
Fair call @dhensby , I have no problem with pedantry  :)  

My editor, Sublime Text, is setup to conform to editor config rules.  Just checking, when I append a comma and hit return the line is indeed padded out with spaces (testing using git diff and vim)

Is the .editorconfig file correct?  I see no mention of tabs, only spaces, and there is no special case for composer.json to be tabbed out.  Running `editorconfig-tools fix composer.json` on the command line also converts the tabs to spaces.


As an aside is there a 'definitive' editorconfig file for SilverStripe modules?

micmania1:
I don't think there is a definitive .editorconfig. We're using PSR-2 and the framework/cms will follow soon. Because we don't specify anything for json files its using spaces which is the default.

If you have time to fix this up (either use tabs or update the rest to spaces) then that would be great otherwise i'd be happy to merge as is and we can fix it up later. @dhensby happy with that?

dhensby:
@gordonbanderson there is a "standard" `.editorconfig` that is used by @helpfulrobot .

Because `composer.json` is typically machine generated I think it's been overlooked, but yes, I suppose it should have a special rule for `composer.json`

We can pull this in and fix it later, but that probably won't happen - I don't see this as a critical PR so would prefer it was right now than someone being bothered to do it later

PR #370, opened 2016-02-04
FIX Success/error message overflow in cms

micmania1:
This is a fix for #210 and #358

### DIFF
--- a/js/cms.js
+++ b/js/cms.js
@@ -2,6 +2,22 @@
 
 	$.entwine('ss', function ($) {
 
+        /**
+         * The page success/error message sits outside of the html block
+         * containing the sidebar and cms fields. This means it overflows
+         * underneath the sidebar.
+         *
+         * @see https://github.com/silverstripe/silverstripe-blog/issues/210
+         */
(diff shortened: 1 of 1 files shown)


js/cms.js
```diff
@@ -2,6 +2,19 @@
...
+        /**
+         * The page success/error message sits outside of the html block
+         * containing the sidebar and cms fields. This means it overflows
+         * underneath the sidebar.
+         *
+         * @see https://github.com/silverstripe/silverstripe-blog/issues/210
+         */
+        $('.cms-content-fields > #Form_EditForm_error').entwine({
+            'onadd': function() {
+                $('.blog-admin-outer').prepend(this);
```
tractorcow: If blog-admin-outer doesn't match anything, does it still move the error? Is the message left safely when, for instance, creating non-blog pages?

micmania1: It _should_ be okay but i'll add a guard in and double check.

tractorcow: Thanks. :)

micmania1:
@tractorcow updated

Issue #271, opened 2015-06-08
Easy to create duplicate Tags & Categories


prembo:
When adding tags, if appears that there the module doesn't check whether the tag exists before adding it.  The problem is exacerbated when the auto-suggest is slow to bring up the list and the user completes the tag entry by hitting enter or typing a comma.

Also maybe a build task to merge duplicate tags would be handy.  Also in the Categories / Tags GridField, the "Move Posts To" dropdown should be ordered alphabetically.  This would assist in trying to move posts from a duplicate tag/category to the proper one, as it is sometimes lost in a sea of unordered tags/categories
![screen shot 2015-06-08 at 1 36 46 pm](https://cloud.githubusercontent.com/assets/1043612/8027860/82075ff8-0de3-11e5-8b5f-72209267d5cc.png)

camfindlay:
+1 @assertchris @tractorcow - came across the same issue when migrating from a 1.0 blog site to 2.0.

prembo:
Here's a Build Task that merges duplicates for anyone who might like to clean up their DB before the the bug is resolved.  Whipped it up fairly quickly so it might be a bit rough, so please back up your DB first:

```php
<?php

/*
 * Build Task which merges duplicate BlogTags
 * @author prembo
 */
class MergeDuplicateBlogTagsTask extends BuildTask{
    protected $title = 'Merge Duplicate BlogTags';
 
    protected $description = 'Merges duplicate BlogTags - relevant to silverstripe-blog >= v2.0';
 
    protected $enabled = true;
 
    function run($request) {
        $this->mergeBlogTags();
    }
    
    private function mergeBlogTags() {
        $tags = BlogTag::get();
        $uniqueTags = [];
        $duplicateTagCount = 0;
        foreach ($tags as $currentTag) {
            if (!array_key_exists($currentTag->Title,$uniqueTags)) {
                $uniqueTags[$currentTag->Title] = $currentTag;
                echo "[$currentTag->ID] $currentTag->Title<br/>";
            } else {
                $duplicateTagCount++;
                $uniqueTag = $uniqueTags[$currentTag->Title];
                echo "DUPLICATE TAG FOUND ---- [$currentTag->ID] $currentTag->Title  ---- is a duplicate of [$uniqueTag->ID] $uniqueTag->Title<br/>";
                $taggedPosts = $currentTag->BlogPosts();
                foreach ($taggedPosts as $taggedPost) {
[...]

PR #372, opened 2016-02-05
FIX Prevent duplicate tags and categories.

micmania1:
This is a fix for #271


micmania1:
I'm not allowed to use integers as ValidationException code... wut? *sigh*

Issue #296, opened 2015-09-14
BUG future publish date doesn't prevent publish


mikeyc7m:
Blog posts with a future publish date are still showing up on the live site. They don't show on the blog listing page but can be accessed directly via their url or in other types of lists.

mikeyc7m:
Looking at the canView function in BlogPost, it seems future posts do not report a PublishDate. I have no idea why it is suppressed.

micmania1:
@mikeyc7m is this still an issue for you?

mikeyc7m:
Yes.

I've installed the latest version, nothing has changed. Future-dated posts are still directly accessible via their URL on a live site, and show in the site search. They are also included in any datalists unless you filter by ClassName.

mikeyc7m:
bump

the publishDate should always be there, perhaps canView could be altered so that when publishDate is missing the function returns false.

dhensby:
Hmm, is this meant to be an embargo feature?

tractorcow:
> I've installed the latest version, nothing has changed. Future-dated posts are still directly accessible via their URL on a live site, and show in the site search. 

This is a bug, from my point of view. These should have canView = false for those without draft permissions.

mikeyc7m:
the new solution seems to work, yay! but... might it be better if the future article URL returned 404 instead of login? not sure what is more relevant.

tractorcow:
In any case, fixed with #373

PR #373, opened 2016-02-05
FIX Prevent showing future blog posts before their publish date

micmania1:
Fix for #296

### DIFF
--- a/code/model/BlogPost.php
+++ b/code/model/BlogPost.php
@@ -460,18 +460,19 @@ class BlogPost extends Page
             return false;
         }
 
+        if($this->canEdit($member)) {
+            return true;
+        }
+
         /**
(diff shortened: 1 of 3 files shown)


code/model/BlogPost.php
```diff
@@ -460,18 +460,19 @@ public function canView($member = null)
...
-        // Show past posts
-        if (!$publishDate->exists() || !$publishDate->InFuture()) {
-            return true;
+        if(!$publishDate->exists()) {
+            return false;
         }
 
-        // Anyone that can edit this page can view it
-        return $this->canEdit($member);
+        return $publishDate->InFuture();
```
tractorcow: You have this backwards; It's visible if the publish date is in the past.

Issue #377, opened 2016-02-16
division by zero error


Jellygnite:
If no tags exist a warning error occurs.

widgets/BlogTagsWidget.php line 93
$tagfactor = 10 / $maxTagCount;

I honestly didn't look into what $tagfactor does but I have changed this line to: 

$tagfactor = 10 / ($maxTagCount > 0 ? $maxTagCount : 1);

Actually after a second look I think it shouldn't even reach this line if tags don't exist. perhaps a check on number of records (tags) found would help.

tractorcow:
I guess you mean BlogTagsCloudWidget.php

$maxTagCount is there are 0 tags, so I guess you can put that whole block into an if($maxTagCount) right?

Jellygnite:
Sorry.  Yes, BlogTagsCloudWidget.php. I copied the wrong filename.

So this 
```
// normalize the tag counts from 1 to 10
           $tagfactor = 10 / $maxTagCount;
            foreach ($tags->getIterator() as $tag) {
                $normalized = round($tagfactor * ($tag->TagCount));
                $tag->NormalizedTag = $normalized;
            }
```
could become

```
if($maxTagCount) {
// normalize the tag counts from 1 to 10
           $tagfactor = 10 / $maxTagCount;
            foreach ($tags->getIterator() as $tag) {
                $normalized = round($tagfactor * ($tag->TagCount));
                $tag->NormalizedTag = $normalized;
            }
}
```

Is that correct?

tractorcow:
Yep.

tractorcow:
Would you like to create a PR with a fix?

gordonbanderson:
@Jellygnite What the tag factor value does is minimize the maximum font size of text in the tag cloud to a sensible level.

PR #381, opened 2016-02-24
FIX: Ensure tag cloud widget does not break when there are zero tags

gordonbanderson:
Fix for my bug reported in #377


dhensby:
Thanks

Issue #350, opened 2015-12-12
GridFieldCategorisationConfig - Performance issues when there are a lot of tags


quamsta:
We have a site with a lot of blog posts and tags (five years' worth of posts with around 3,000 tags!) we imported from Wordpress.

In the Blog class, the GridFieldCategorisationConfig component appears to load a "move to" dropdown that's populated with every tag in the db for each entry in GridField. Apparently $this->Tags() and $this->Categories() is called for each individual tag/category entry in the GridField to populate the dropdowns. This pegs our CPU at 99% for a long time and causes loading the Blog page in SilverStripe to be very slow.

Removing the GridFieldCategorisationConfig component from the Tags/Categories field in Blog speeds up the blog drastically. Caching all of the tags/categories first and loading them for all subsequent dropdowns might help, or somehow lazily loading the "Move to" dropdowns might help too.

dhensby:
I'm not sure how 3,000 tags can be helpful for a blog reader are they really meant to look through 3,000 tags to find the tags they are interested in?

the Blog clearly isn't designed to work at that scale and I doubt it's a priority as 3k tags is pretty nuts.

Efficiency PRs will be accepted but I don't think it'll be a core focus for some time

quamsta:
Just to make sure we're clear: this is 3,000 total tags from over 5+ years and 2,000 posts from a large music review and news blog. The issue is that when we view the Blog, the GridField populates each row with a dropdown that has a massive query for tags. 

What I'm hearing is that the blog module is not scalable to a degree that's useful for larger, longstanding projects and that "just don't have so much content" is the solution here. 

Not really very forward-thinking for a *content* management system, don't you think?

stevie-mayhew:
An option is to replace those tag calls with the (silverstripe-select2)[https://github.com/sheadawson/silverstripe-select2] module. We use this in instances where there is a large number of things to be shown in a dropdown, such as members.

@dhensby would this be a good addition here? I think that with the work going on for 4.0 in FE that chosen will likely be moved out and a better system put in place, but for now this does fix the issue and I think is a good improvement. 

Its not a supported module though, so there might be some issues there.

dhensby:
> this is 3,000 total tags from over 5+ years and 2,000 posts from a large music review and news blog

[...]

PR #383, opened 2016-03-04
FIX Performance issues when a blog has a large number of tags.

micmania1:


code/admin/GridFieldMergeAction.php
```diff
@@ -69,27 +89,48 @@ public function getColumnsHandled($gridField)
...
-
-            return $dropdown->Field() . $action->Field() . '<a title="Move posts to" class="MergeActionReveal">move posts to</a>';
+        if ($columnName === 'MergeAction') {
+
+            if(!$record->hasMethod($this->childMethod) || $record->{$this->childMethod}()->count() < 1) {
+                return '';
+            }
+
+            $children = $record->{$this->childMethod}();
+            $cacheKey = md5($children->dataQuery()->sql());
```
tractorcow: Won't work if two different queries have the same SQL, but different parameters. :)

tractorcow: What parameters do you need... $record->ID and $this->childMethod right? That might be enough.

tractorcow: Please update cacheKey to include hashed params. You can get them passed by by ref from the sql($args) method.

dhensby:
Bump

micmania1:
This didn't help so i'm going to close. I'll follow up on the original ticket #350

PR #400, opened 2016-05-30
Tidied the badges on reademe

micmania1:
Its looking a bit messy at the moment and there's a lot of useless badges.

Thinking is:
First line: code quality
Second line: informational

See the readme here: https://github.com/micmania1/silverstripe-blog/tree/tidy-readme


micmania1:
Needless to say, but the test failures aren't related :)

PR #401, opened 2016-05-30
FIX closing p tag

micmania1:
Carried on from #343

### DIFF
--- a/templates/Includes/PostSummary.ss
+++ b/templates/Includes/PostSummary.ss
@@ -13,14 +13,15 @@
 	</p>
 
 	<% if $Summary %>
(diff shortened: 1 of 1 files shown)


tractorcow:
Sorry to get to this so late, but how is the p tag not closed properly? the if-else always opens the p tag, and the following html block always closes it.

micmania1:
I think they did close but the HTML was incorrect. `$Summary` added its own p tags so we basically had ```
<p>
    <p>$Summary</p> 
    $Content
</p>
```

tractorcow:
Oh, that makes sense!

PR #403, opened 2016-05-31
FIX #394 add getDate to proxy publish date in rss template

### LINKED ISSUE #394: PubDate in RSS feed is wrong
The date that ends up in the RSS feed is the post's creation date, and not its publish date. I've worked around this by creating a custom RSSFeed.ss template that replaces:
			<% if $Date %><pubDate>$Date.Rfc822</pubDate>
			<% else %><pubDate>$Created.Rfc822</pubDate><% end_if %>

 with the following:
			<% if $PublishDate %><pubDate>$PublishDate.Rfc822</pubDate>
			<% else_if $Date %><pubDate>$Date.Rfc822</pubDate>
			<% else %><pubDate>$Created.Rfc822</pubDate><% end_if %>

This works, but it would be better if it worked correctly out of the box.

micmania1:


dhensby:
thanks

PR #404, opened 2016-05-31
Add canEdit extension

samwalshnz:


micmania1:
@samwalshnz thanks for contributing but there should already be an extension point which is inherited from SiteTree. See http://api.silverstripe.org/3.3/source-class-SiteTree.html#1052 

I'm going to close this but if there's an issue with the existing implementation feel free to comment and i'll re-open.

samwalshnz:
@micmania1 SiteTree's canEdit method gets called, but only if it returns `true` does BlogPost return `true`. But if you wanted to extend BlogPost's canEdit, then a `false` response gets ignored. https://github.com/silverstripe/silverstripe-blog/blob/master/code/model/BlogPost.php#L521

micmania1:
The only time the extension wouldn't be called is if you're logged in with either `ADMIN` or `SITETREE_EDIT_ALL` permissions which I think is reasonable. What's your use case?

samwalshnz:
@micmania1 My usage case is that some of the blog posts are synchronised in from another source. Anything changed on a Blog Post will be overwritten on the next sync, so for these specific blog posts, I don't want anyone (even admin) editing them. The extension still allows defaulting to checking if admin or not if the developer choses to do so.

micmania1:
You could extend `BlogPost` with `ImportedBlogPost` and overload the `canEdit` method to return false. We can't add another extension point here because it would execute all extension points twice.

PR #406, opened 2016-06-10
Localisation for multiple authors

TheBnl:
added localisation for multiple authors in the entry meta


tractorcow:
Can you please also add these translation strings to https://www.transifex.com/silverstripe/silverstripe-blog/? Otherwise, next release these strings will get deleted. :) Updates are only one way, not between the two repos, unfortunately.

Thank you very much!

Issue #382, opened 2016-02-29
RSS feeds for Categories and Tags


sekjal:
RSS feeds only work for an entire Blog at this time; the only way then to have multiple feeds based on interest is to create multiple blogs.  Having tag and category specific feeds would help users subscribe to only the content they want to see, and increase the possibilities for automation using third-party tools like IFTTT or Zapier.

micmania1:
This would be a welcomed feature. Do you have the time to create a PR?

sekjal:
Yes, I should within the next few weeks.

hdrlab:
Yes, this would be a welcome feature. I was just searching for how to get an RSS feed for categories. My goal is to allow users to signup for email updates for specific topics (categories). Per-category RSS feeds would be ideal for allowing each topic to get the right articles.

tractorcow:
You could update the `rss` action to allow `category` or `tag` to be assigned the `$ID` param, where the urlsegment of the category or tag could be the `$OtherID` parameter.

@sekjal are you looking closer to this pull request?

sekjal:
Unfortunately, I've been pulled away in the last couple months on other urgent projects.  I'm hoping to get to this in June, but if anyone else wants to take a stab at it first, I'd be happy to test.

micmania1:
Added in #407

PR #407, opened 2016-06-13
Add RSS feeds for categories and tags

ksdhans:
This patch adds RSS feeds for tags and categories (for issue #382). The RSS feeds are at the following URLs:
- */category/<category-name>/rss
- */tag/<tag-name>/rss

### DIFF
--- a/code/model/Blog.php
+++ b/code/model/Blog.php
@@ -644,8 +644,8 @@ class Blog_Controller extends Page_Controller
      * @var array
      */
     private static $url_handlers = array(
-        'tag/$Tag!' => 'tag',
-        'category/$Category!' => 'category',
+        'tag/$Tag!/$Rss' => 'tag',
+        'category/$Category!/$Rss' => 'category',
         'archive/$Year!/$Month/$Day' => 'archive',
         'profile/$URLSegment!' => 'profile',
     );
@@ -834,7 +834,12 @@ class Blog_Controller extends Page_Controller
 
         if ($tag) {
             $this->blogPosts = $tag->BlogPosts();
-            return $this->render();
+
+            if($this->isRSS()) {
+            	return $this->rssFeed($this->blogPosts);
+            } else {
+            	return $this->render();
+            }
         }
 
         $this->httpError(404, 'Not Found');
@@ -874,7 +879,11 @@ class Blog_Controller extends Page_Controller
         if ($category) {
             $this->blogPosts = $category->BlogPosts();
 
-            return $this->render();
+            if($this->isRSS()) {
+            	return $this->rssFeed($this->blogPosts);
+            } else {
+            	return $this->render();
+            }
         }
 
         $this->httpError(404, 'Not Found');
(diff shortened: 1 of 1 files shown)


code/model/Blog.php
```diff
@@ -1093,4 +1098,30 @@ public function getRSSLink()
...
     }
+    
+    /**
+     * Displays an RSS feed of the given blog posts.
+     *
+     * @return string
+     */
+    public function rssFeed($blogPosts)
+    {
+        $rss = new RSSFeed($blogPosts, $this->Link() . 'category/general/', $this->MetaTitle, $this->MetaDescription);
```
micmania1: This still has hard-coded url segments.

code/model/Blog.php
```diff
@@ -1093,4 +1098,30 @@ public function getRSSLink()
...
+
+        $this->extend('updateRss', $rss);
+
+        return $rss->outputToBrowser();
+    }
+    
+    /** 
+     * Returns true if the $Rss sub-action for categories/tags has been set to "rss"
+     */
+     private function isRSS() {
```
micmania1: Could you drop the bracket to the next line please (psr-2) and document the return type?

code/model/Blog.php
```diff
@@ -1093,4 +1098,30 @@ public function getRSSLink()
...
+     	 $rss = $this->request->param('Rss');
+     	 if($rss && strcasecmp($rss, "rss") == 0) {
```
micmania1: Could you change `if ($rss ...` to `if(is_string($rss)...`. strcasesmp *can* throw an error in some circumstances - this is just a little extra safety.

code/model/Blog.php
```diff
@@ -1093,4 +1098,30 @@ public function getRSSLink()
     {
         return $this->Link('rss');
     }
+    
+    /**
+     * Displays an RSS feed of the given blog posts.
+     *
+     * @return string
+     */
+    public function rssFeed($blogPosts)
```
micmania1: Could you make this protected please?

micmania1:
Thanks for the contribution @ksdhans.

I've left a few small comments but other than that i'd be happy with this being merged.

In future we'd probably like to split out tags and categories to their own controller with their own `rss` method as to avoid nasty method names and also the `isRss` check which feels a little hacky. (we could have `isJson`, `isXml` etc.) which would eventuate in quite nasty if statements.

Its good to see progress and marking `isRssFeed` as private is a good idea - this should allow us to make the above changes without any BC issues in future 👍 .

I'll wait for any other feedback and your updates before merging.

ksdhans:
Okay, I've made the changes. However, I've just discovered that the RSS feed's link/atom:link always points to the root URL. So it's not quite right yet.

ksdhans:
The link URLs are now correct for category/tag rss feeds, so I think that it's all working as it should.

code/model/Blog.php
```diff
@@ -1093,4 +1098,34 @@ public function getRSSLink()
...
+
+        return $rss->outputToBrowser();
+    }
+    
+    /** 
+     * Returns true if the $Rss sub-action for categories/tags has been set to "rss"
+     */
+     private function isRSS() 
+     {
+     	 $rss = $this->request->param('Rss');
```
micmania1: The formatting seems a bit strange here (tabs/spaces) - not really sure what's going on.

micmania1:
@ksdhans just one small thing.

@silverstripe/module-maintainers-blog-comments-and-related-modules anybody got any input before merging?

code/model/Blog.php
ksdhans: Yeah, that's the code editor that I'm using being weird (and me not noticing). Sorry about that.

micmania1:
@ksdhans merged! Thanks 👍

PR #411, opened 2016-07-15
remove i18n formatting in links

lekoala:
the getArchiveMonth and getArchiveYear only accepts number in Blog.php. Removing i18n formatting restore the functionnality that is currently not working.


dhensby:
heh - seems sensible :P

PR #414, opened 2016-09-08
FIX Excerpt should be casted as HTMLText

### LINKED ISSUE #413: Double encoded html entities in excerpt
It seems that & character (possible others, but not all entities) is double encoded (`&amp;amp;`) in the blog post excerpt. It works in summary and content, but not in the excerpt generated from content.

dhensby:
fixes #413

### DIFF
--- a/code/model/BlogPost.php
+++ b/code/model/BlogPost.php
@@ -83,7 +83,7 @@ class BlogPost extends Page
      * @var array
      */
     private static $casting = array(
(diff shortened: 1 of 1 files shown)


chillu:
Given that used the `HTMLText->Summary()` method is casted the same, this makes sense to me.

dhensby:
THANKS - damn that code coverage takes FOREVER!

Issue #405, opened 2016-06-01
SilverStripe 4.x compatibility


micmania1:
I've started work on this here: https://github.com/micmania1/silverstripe-blog/tree/ss4
- [x] API updated
- [x] Tests passing
- [ ] Update composer (including dependencies - use ss4 branches for now where available)
- [x] Move shared Tag/Category/URLSegment functionality into a trait and share on BlogTag and BlogCategory
- [x] Remove compass (in favour of gulp) for better future-proofing
- [ ] Test/Update to use SS4s new theme system
- [ ] Test/Update to ensure it works with SS4 namespace changes
- [ ] Remove code for old PHP version (eg. https://github.com/micmania1/silverstripe-blog/blob/ss4/code/model/Blog.php#L116)

Bugs:
- [x] Right sidebar - floats left instead of right (also hides bottom right preview options)
- [x] Right sidebar - update open/close icon at the bottom to float left when open
- [x] Tags & Categories listboxes not working
- [x] Add new blog post above gridfield takes you to 404 page

Blockers:
- [x] https://github.com/silverstripe/silverstripe-framework/issues/5625

Further Investigation/Ideas/Not Required:
- [ ] Should we implement namespacing in Blog? What effect will this have on upgrades?
- [ ] Should we remove migration scripts from v1 to v2 when v3 is released?
- [ ] Consider ways to extend page (see https://github.com/silverstripe/silverstripe-framework/issues/5844)

micmania1:
ps. Can someone add a 3.x label please? It allows me to add milestones but not labels :/

micmania1:
Branches where this work lives:
Blog: https://github.com/micmania1/silverstripe-blog/tree/ss4
Lumberjack: https://github.com/micmania1/silverstripe-lumberjack/tree/ss4
Tagfield: https://github.com/micmania1/silverstripe-tagfield/tree/ss4

tractorcow:
Good job @micmania1 :D Sorry to break everything, but remember to update for namespacing. Ping me if you need help running the upgrade tool.

sminnee:
@micmania1 can you:

 * raise PRs for your work in progress
 * put the to-do items 
 * give me write access to your micmania1 forks so I can push work there?

robbieaverill:
Hey @micmania1, @sminnee and @tractorcow  - I'm working on this as part of a project upgrade at the moment. I've forked of from @micmania1's initial work - here's my status with blog and its dependencies:

## Completed

[...]

PR #421, opened 2016-12-20
SilverStripe 4 compatibility

robbieaverill:
## Note

Carries on from work done by @micmania1  in #405.

## Summary

Updated for compatibility with SilverStripe 4 in its current state. There are still some things to be done, but this is largely working to start off with. Brief summary of changes:

* Composer: dev alias added, PSR-4 autoloading compatibility
* Structure: "code" -> "src", namespacing added and implemented
* Travis: config updated for SS4 (PHP 7.x is no longer allowed to fail)
* PSR-2: some style formatting changes
* Asset building: Gulp used instead of compass

## Blockers

- [x] silverstripe/silverstripe-tagfield#87
- [x] silverstripe/silverstripe-lumberjack#48
- [x] silverstripe/silverstripe-comments#194
- [x] silverstripe/silverstripe-widgets#143
- [x] colymba/GridFieldBulkEditingTools#136

## Considerations

- [x] Do we need to create/update branches other than master before it becomes SS4 only?


dhensby:
OK - master is now aliased as `3.x-dev` and is where we should be putting SS 4 changes. please make sure to update `.travis.yml` and constraints in `composer.json`

robbieaverill:
@dhensby I've rebased and pushed. If you want me to squash a few of these commits, happy to do so

micmania1:
Probably more a question for @sminnee / @chillu / @tractorcow but what's the guidance around committing the .upgrade.yml file? Is it useful after you've ran the upgrade tool?

@robbieaverill haven't tested this but looks good 👍 Glad to see those migration tasks gone :D

robbieaverill:
@micmania1 it's useful in that it tells other modules that are being upgraded what the namespaces to apply for Blog classes are. Without the mapping they'll all just be imported into the global namespace when you upgrade a module or project that uses this module, e.g. `use BlogPost;` as opposed to `use SilverStripe\Blog\Model\BlogPost;`

robbieaverill:
Once the next alpha release of framework comes out this PR or another will need to be updated to extend `PageController` instead of `Page_Controller`. Doing it now would probably fail builds by pulling in the latest alpha release

src/Model/BlogTag.php
```diff
@@ -64,7 +64,7 @@ class BlogTag extends DataObject implements CategorisationObject
      */
     protected function getListUrlSegment()
     {
-        return 'tags';
+        return 'tag';
```
micmania1: Was this meant to be changed? It should match current functionality.

robbieaverill:
Yeah, this changes it back to current functionality

dhensby:
Looks like the comments module needs upgrading?

robbieaverill:
Yeah it does, and there is some widget work to do too. I'll be looking at this early next week.

dhensby:
Ok, then they are another couple of blockers to list

robbieaverill:
Recent discussion - should we move the comments and widgets code into a separate module and suggest that instead, e.g. `silverstripe/blog-comments` and `silverstripe/blog-widgets` ?

sminnee:
> Recent discussion - should we move the comments and widgets code into a separate module and suggest that instead, e.g. silverstripe/blog-comments and silverstripe/blog-widgets ?

I'd suggest getting an SS4 version merged first and then look at further refactorings. Otherwise this will becoming a zombie PR.

robbieaverill:
@sminnee I just reached for the zombie emoji response - sadly disappointed.

tractorcow:
Testing this locally:
 
 - [ ] categories and tags aren't saving

tractorcow:
There's a test failing; I'll push a fix if you give me write access. :)

robbieaverill:
@tractorcow I've fixed the tests, I had to update a few things:

* There were parts where I'd missed namespacing the fixture classes in the tests, and some parts (blog widget classes) where I seem to have completely forgotten to implement namespaces at all
* [This multibyte logic](https://github.com/silverstripe/silverstripe-blog/blob/2.4.5/code/extensions/URLSegmentExtension.php#L39-L43) got lost somehow, re-introduced
* [This logic](https://github.com/silverstripe/silverstripe-blog/blob/2.4.5/code/extensions/URLSegmentExtension.php#L23) was changed to `if (empty($this->URLSegment))` during the early refactor and move the the BlogObject trait. I've added the empty check back, checking second for the URL Segment being empty - this matches existing behaviour (v2.4.5)
* I've switched out other places where FQCNs were string literals for `BlogPost::class` equivalents with imports at the top
* I added namespaced to the tests while I was in there

---

Another thing I noticed was that lumberjack isn't removing blog posts from the SiteTree when they're displayed in the GridField. Logged as silverstripe/silverstripe-lumberjack#50

robbieaverill:
@dhensby - we're green, finally 💚

dhensby:
43 mins is still far too long ;)

PR #426, opened 2017-02-07
setValue on lazy dbObjects

andrewandante:
Currently, unless you are logged in you can't see a BlogPost. This is because `$this->dbObject('PublishDate')` doesn't auto-populate the value, as it picks up `PublishDate_Lazy` instead of `PublishDate`.

This change sets the value explicitly to ensure that setting view permissions works as expected.


dhensby:
We should aim to fix this in core rather than adding a workaround to blog module

andrewandante:
Being fixed here: https://github.com/silverstripe/silverstripe-framework/pull/6604

Issue #427, opened 2017-02-17
Dev/Build throws user warnings on localisation


andrewandante:
On latest master (dev-master 92004f8), running dev/build throw the following `[User Warnings]`:

- Localisation entity en.BlogCategoriesWidget.Direction is invalid
- Localisation entity en.BlogCategoriesWidget.Limit is invalid
- Localisation entity en.BlogCategoriesWidget.Sort is invalid
- Localisation entity en.BlogTagsWidget.Direction is invalid
- Localisation entity en.BlogTagsWidget.Limit is invalid
- Localisation entity en.BlogTagsWidget.Sort is invalid
- Localisation entity en.BlogCategoriesWidget.Direction is invalid
- Localisation entity en.BlogCategoriesWidget.Limit is invalid
- Localisation entity en.BlogCategoriesWidget.Sort is invalid
- Localisation entity en.BlogTagsWidget.Direction is invalid
- Localisation entity en.BlogTagsWidget.Limit is invalid
- Localisation entity en.BlogTagsWidget.Sort is invalid

robbieaverill:
This is likely to be from the i18n changes in the framework recently. Updating legacy syntax is simple but tedious - I'll have a go at it in the next day or two if I have some time.

andrewandante:
Is there a guide? Happy to do it myself. Simple but tedious I can do :p

robbieaverill:
@andrewandante actually this isn't what I thought (although still related to the i18n upgrade) - looks like all those translations are nested three levels (see `blog/lang/en.yml` `en.BlogCategoriesWidget.Direction`). I'm assuming the new i18n backend doesn't support this level of nesting, so we'll need to flatten this out a little bit.

dhensby:
Isn't that exactly how it should be nested?

robbieaverill:
Haha, there's another key I didn't include in that example. Basically these translations are arrays where the majority of translations are strings at that nesting level.

dhensby:
Ah. Ok

tractorcow:
Sorry, ss 4.0 does support nested strings for localisation, but only for specific purposes (e.g. pluralisation, which is localised as an associative array).

We need to un-nest these, and maybe use a `_` in place of `.` for those keys.

PR #428, opened 2017-02-17
showcalendar config method change

andrewandante:
Old `setConfig()` method no longer exists


andrewandante:
Test failures are because of #427

PR #433, opened 2017-04-02
Update UploadField to one that exists

andrewandante:


dhensby:
We need to remove the 5.5 test run from .travis.yml as well as renaming a bunch of namespaces that have now been split out

dhensby:
I suppose this is a step in the right direction

PR #434, opened 2017-04-06
Allow other checks on canView for blog

hailwood:


micmania1:
@hailwood This already gets called in `parent::canView()`. What's your use-case/issue?

hailwood:
Hmm you're right; Originally when I tried it I must have done something wrong as it wasn't getting called so thought it was missed. Closing.

PR #437, opened 2017-05-03
add meta links for previous and next pages in paging and post view

xini:
this helps search engines to find the pager and index more than just the first page of posts

### DIFF
--- a/code/model/Blog.php
+++ b/code/model/Blog.php
@@ -758,7 +758,6 @@ class Blog_Controller extends Page_Controller
 
         if ($year) {
             $this->blogPosts = $dataRecord->getArchivedBlogPosts($year, $month, $day);
-
             return $this->render();
         }
 
@@ -838,7 +837,7 @@ class Blog_Controller extends Page_Controller
             if($this->isRSS()) {
             	return $this->rssFeed($this->blogPosts, $tag->getLink());
             } else {
-            	return $this->render();
+                return $this->render();
             }
         }
 
@@ -882,7 +881,7 @@ class Blog_Controller extends Page_Controller
             if($this->isRSS()) {
             	return $this->rssFeed($this->blogPosts, $category->getLink());
             } else {
-            	return $this->render();
+                return $this->render();
             }
         }
 
@@ -1043,7 +1042,37 @@ class Blog_Controller extends Page_Controller
 
         return $posts;
     }
+    
+    /**
+     * Returns the absolute link to the next page for use in the page meta tags. This helps search engines
+     * find the pagination and index all pages properly.
+     * 
+     * @example "<% if $PaginationAbsoluteNextLink %><link rel="next" href="$PaginationAbsoluteNextLink"><% end_if %>"
+     * 
+     * @return string
+     */
+    public function PaginationAbsoluteNextLink() {
+        $posts = $this->PaginatedList();
+        if ($posts->NotLastPage()) {
+            return Director::absoluteURL($posts->NextLink());
+        }
+    }
 
+    /**
+     * Returns the absolute link to the previous page for use in the page meta tags. This helps search engines
+     * find the pagination and index all pages properly.
+     * 
+     * @example "<% if $PaginationAbsolutePrevLink %><link rel="prev" href="$PaginationAbsolutePrevLink"><% end_if %>"
+     *
+     * @return string
+     */
+    public function PaginationAbsolutePrevLink() {
+        $posts = $this->PaginatedList();
+        if ($posts->NotFirstPage()) {
+            return Director::absoluteURL($posts->PrevLink());
+        }
+    }
+    
     /**
      * Displays an RSS feed of blog posts.
      *
@@ -1128,4 +1157,5 @@ class Blog_Controller extends Page_Controller
             return false;
         }
     }
+	
 }


robbieaverill:
Hey @xini - thanks for the PR! This is a feature, so would need to be targetted at the `2` branch for the next minor release (2.5.0). Any chance of getting some unit tests for this too?

xini:
Hi @robbieaverill,
How would you test this? 
Also, I was wondering whether this should actually go into PaginatedList to have the functionality available not just in the blog?
Cheers

robbieaverill:
@xini I haven't looked very thoroughly at the context for the change, but if you want to test this you can use `FunctionalTest` to retrieve the HTML from a controller that uses these methods and assert that the correct meta tags are in the result

code/model/BlogPost.php
```diff
@@ -715,4 +715,37 @@ protected function onBeforeWrite()
...
+			Requirements::insertHeadTags('<link rel="next" href="' . Director::absoluteURL($post->Link()) . '" />', 'nextlink');
+		}
+		if ($post = $this->getPreviousPost()) {
+			Requirements::insertHeadTags('<link rel="prev" href="' . Director::absoluteURL($post->Link()) . '" />', 'prevlink');
+		}
+	}
+	 
+	public function getPreviousPost()
+	{
+		return CustomBlogPost::get()->filter(array(
```
micmania1: You should use `$this->Parent()` rather than manually building the query

code/model/BlogPost.php
```diff
@@ -715,4 +715,37 @@ protected function onBeforeWrite()
...
+		))->sort(array(
+			"PublishDate" => "DESC",
+		))->first();
+	}
+	
+	public function getNextPost()
+	{
+		return CustomBlogPost::get()->filter(array(
+			"ParentID" => $this->ParentID,
+			"PublishDate:GreaterThan" => $this->dbObject('PublishDate')->Rfc2822(),
```
micmania1: It should be `GreatThanOrEqual` and exclude the current post id. For example, If a script automatically pulled in blog posts from an external source, then blog posts may be created and published at the same time.

code/model/BlogPost.php
```diff
@@ -715,4 +715,37 @@ protected function onBeforeWrite()
...
+	{
+		parent::init();
+		 
+		// check pages and add links
+		if ($post = $this->getNextPost()) {
+			Requirements::insertHeadTags('<link rel="next" href="' . Director::absoluteURL($post->Link()) . '" />', 'nextlink');
+		}
+		if ($post = $this->getPreviousPost()) {
+			Requirements::insertHeadTags('<link rel="prev" href="' . Director::absoluteURL($post->Link()) . '" />', 'prevlink');
+		}
```
micmania1: I feel like this should be in the template rather than here. We don't inject HTML anywhere else in the module and it seems like a "old" way over doing things. We should provide a good API that makes it easy to do this though. For example:

```html
<% with $PrevBlogPost %>
    <link rel="next" href="$Link" />
<% end_if %>
```

xini: ok, no problem.

code/model/BlogPost.php
xini: You're right. but how would I get the previous post from the PaginatedList of the Parent?

xini:
Hi @micmania1,
- you're right, browsing within tags and categories is covered on blog level, but not on blog post level.
- I think this should not be in a module but in blog core. I was told Google needs these links to properly find the pagination.
- I was thinking about moving the functionality into PaginatedList (making that implement GlobalTemplateProvider) to make the links available to other modules as well? e.g. with $PaginationPreviousLink and $PaginationNextLink. But what about multiple paginated lists on a page?
Cheers

code/model/BlogPost.php
micmania1: Its quite difficult - the only way I can think of is to load the entire list then figure it out, or write a horrible custom query.

I guess we could be opinionated here and say the sort is by publish date, regardless of what the user/dev sets. Maybe if we can add an extension point so this functionality can be changed.

eg.
```php
$prevQuery = $this->Parent()->filter(/* some stuff */)->sort('blah');
$this->extend('updatePrevPostQuery', $prevQuery);
return $prevQuery->first():
```

micmania1: This shouldn't depend on pagination though as blog post doesn't know anything about pagination.

micmania1:
@xini Maybe for this PR we could just add the next/prev links for the paginated page? It seems quite solid and useful. 

There's a few things which make next/prev blog post a bit difficult so this could be tackled separately. It can be applied as a `DataExtension` if you need it locally in the meantime.

code/model/BlogPost.php
xini: By default the posts are sorted by publish date and this can be changed via hook in Blog::getBlogPosts(). 
You're right, pagination shouldnt be involved here. 
The correct way of doing it would be to use Parent()->getBlogPosts(), find the position of the current post and then go back or forth. But how would we incorporate navigation within a category or tag here? If we go from the post to the Parent(), that doesn't know anything about the filter that was applied...

xini:
@micmania1 sounds reasonable. what do you think about the idea to add it in the PaginatedList class instead?

micmania1:
@xini its probably already possible actually...

```
<% with $Pagination %>
    <% if $NotLastPage %>
        <link rel="next" href="$NextLink" />
    <% end_If %>
<% end_with %>
```

xini:
@micmania1 unfortunately not. And I can't find a way to do it globally. So I'll just clean up this PR...

dhensby:
Todo before merge: target 2 branch and rebase

xini:
@dhensby done, thanks.

xini:
@micmania1 @dhensby can this be merged please? When can I expect a release with this? thanks.

dhensby:
@micmania1 I'd like to have you re-review this, please :)

xini:
any updates on this?

xini:
ping...! ;)

xini:
@micmania1 can you please review this? thanks.

dhensby:
OK - I have concerns about the repeated calling of `$this->PaginatedList()` as I think that'll be pretty expensive in terms of CPU usage...

Is there any way to make this more efficient?

xini:
@dhensby thanks for the review. `$this->PaginatedList()` only creates the pagination wrapper around `$this->blogPosts` which are already loaded. To my knowledge it shouldn't trigger any database queries, right? I don't see how I could skip accessing `$this->PaginatedList()`, apart from storing it in another protected var `$this->paginatedList`. Would you prefer that?

dhensby:
> $this->PaginatedList() only creates the pagination wrapper around $this->blogPosts which are already loaded.

Yes, you're right - I suppose the overhead isn't so bad, then

dhensby:
> Did you consider if the user is browsing a tag or category and they click on a blog post, should the 'Next' link take into account the tag or category they've just came from? If devs start to use the next/prev functions to produce links as is common in wordpress, then it might be a bit strange for a user to be taken to completely separate category when clicking next.

Has this been addressed?

xini:
@dhensby it has been addressed in the way that the prev/next links are only available on the overview pages, not the individual blog posts. The main intention for this change was binding the pagination together for search engines (see https://support.google.com/webmasters/answer/1663744?hl=en). the prev/next links within posts are an additional feature that can be addressed later if needed. For that we would probably have to set the main/category/tag in the session in order to figure out where the user was. I'm not a huge fan of that though, as it neds a cookie which messes with cacheability in Varnish.

dhensby:
ok, lets get this in

xini:
thanks, mate.

PR #438, opened 2017-05-04
FIX: updated test suite to cover supported versions

sminnee:


sminnee:
This will fail until https://github.com/silverstripe/silverstripe-framework/pull/6872 is merged but I'm hoping that this refactoring will make it _possible_ to run the coverage tests. ;-)

sminnee:
OK — have amended. It looks like the original fixture was relying on undocumented behaviour to fix inconsistencies in the fixture ;-)

tests/blog.yml
```diff
@@ -1,6 +1,7 @@
 # Mock date is set to 2013-10-01 20:00:00
+# Write to tables rather than dataobjects for performance
```
robbieaverill: Hmm - is this what we're going to encourage? I don't know if I like this to be honest

sminnee: You have a choice: do this or don't get coverage tests. In addition, it speeds up tests 50x.

I think it's worth it.

robbieaverill: I must've missed the coverage tests issue. This is going to get super confusing for people if we encourage it over model names given that the default behaviour for DOs is to use the FQCN as the table name?

sminnee: Fixtures are always crafted to suit a test-case — the fact that DataObject::write() is called doesn't magically make the test declarative.

In the example of porting this test over, the one issue that I came across was that the write process automatically changed this:

```yml
BlogTag:
  Title: cat
  URLSegment: cat-tag
```

To this

```yml
BlogTag:
  Title: cat
  URLSegment: cat
```

So when I wrote the YML I had to write the latter, not the former.

Frankly I think this is an improvement.

sminnee:
Only failure now is codecov complaining about previous data being missing.

tests/blog.yml
sminnee: > This is going to get super confusing for people if we encourage it over model names given that the default behaviour for DOs is to use the FQCN as the table name?

It doesn't seem to be in this module, and I think that the default table name generation is under review.

robbieaverill: Cool - fair points

sminnee:
@robbieaverill I think the long-term fix is to move away from bloaty use-everywhere fixtures and instead scaffold on a per-test basis the minimum content necessary to run the test in question.

however that's a bigger change.

PR #444, opened 2017-06-14
Prevent user_error when querying object from Solr

alex-dna:


dhensby:
Merging as failures are:

1. Unable to generage coverage report
2. Unable to test hhvm

PR #447, opened 2017-06-20
give BlogPost ownership of FeaturedImage

### LINKED ISSUE #429: Versioned assets break established publish behaviour
With the addition of versioned assets, publishing a blog post is not as simple as it once was. For example, as a content author, I would assume that when I publish my blog post with a feature image, the image itself is also published - however, it remains in draft.

The Campaigns feature seems designed to solve this, so we could look into hooking each BlogPost into a small campaign that includes attached images, presumably with some config to toggle this behaviour.

Alternatively, a BetterButton that warns onPublish that assets foo and bar will also be published could be an option.

This wouldn't be as annoying if you could push the asset from stage to live from within the BlogPost itself, but you have to navigate into the asset-admin, find the asset and publish it manually.

Other solutions welcomed with warm and loving arms.

andrewandante:
Fixes #429 

Will publish the FeaturedImage alongside the BlogPost.

### DIFF
--- a/src/Model/BlogPost.php
+++ b/src/Model/BlogPost.php
@@ -75,6 +75,13 @@ class BlogPost extends Page
     /**
      * @var array
      */
+    private static $owns = array(
+        'FeaturedImage',
+    );
+
+    /**
+     * @var array
+     */
     private static $many_many = array(
         'Categories' => BlogCategory::class,
         'Tags'       => BlogTag::class,


dhensby:
What's up with the test failures?

robbieaverill:
Travis config updated at #459

andrewandante:
Ah - will rebase once merged :)

dhensby:
Merged

andrewandante:
Now we passin' @dhensby

PR #451, opened 2017-07-04
total refactor of blog.yml for tests, use classes properly

andrewandante:
Rework fixture to fit standard scheme, use proper classes, other fixes


andrewandante:
@dhensby kick this back off when you've fixed Travis please :)

PR #463, opened 2017-08-24
Small fixes in Dutch translation

JorisDebonnet:
Fixed a glaring typo (Auterus => Auteurs), and a few more minor things


robbieaverill:
Hey @JorisDebonnet - you'll need to make this change in Transifex which is the source of truth for non-English translations. If this was merged it would simply be overwritten the next time we pull down from Transifex. See [here](https://www.transifex.com/silverstripe/silverstripe-blog/translate/#nl/master-v2/46454955) for the place you'll need to make these changes. Thanks!

Issue #376, opened 2016-02-10
Non-english tags/categories can create empty URLSegments


micmania1:
Create a tag with the title ` شخصی و افراد` and save.
Click the tag in the front-end - notice there is no URLSegment and the link leads to 404.

Disclaimer: I have no idea what the example tag says or even what language it is.

tractorcow:
I think this probably needs fixing over at https://github.com/assertchris/silverstripe-segment-field

alialamshahi:
Thanks micmania1,
The language is Farsi, both tags and categories URLSegment don't support non-English letters. I updated database manually but every time I assign a post with a tag or category their value becomes NULL again!

micmania1:
@tractorcow we don't use that in blog. Or do we?

Can anybody tell me how the CMS behaves in this scenario?

tractorcow:
The normal CMS Is theoretically ment to support extended characters, if URLSegmentFilter.default_allow_multibyte is on. However, this is for pages only, not tags / categories.

It would be good if segment-field extension also followed this rule. ;)

alialamshahi:
Any updates on this issue? I'm still editing URLSegments in the database manually (through PHPadmin), every time a category or a tag are used in a blog post their URLSegments becomes NULL again!

micmania1:
@alialamshahi can you check out #402 and let me know if this resolves your issue?

alialamshahi:
@micmania1 works just fine! thanks a million!

Misagh90:
Hi,
@micmania1 @alialamshahi 
It's happen even in blog_profile.ss for Author, the url goes to 404 page in Persian (Farsi).

PR #466, opened 2017-09-01
FIX Encode URLSegment to support multibyte member profile URLs

### LINKED ISSUE #465: Blog Profile does not support UTF-8
Hi,
I create a yml file in my config & i add
```
URLSegmentFilter:
  default_allow_multibyte: true
```
to create Persian(Farsi) URL link but it's not working on blog Profile pages. (it's work on another kind of Pages) (I try in another languages except English, but still not working)
I thinks that not supporting UTF-8.
how can i fix this?!

robbieaverill:
This changes:

* The length of the URLSegment DB column, since 50 characters is not enough and some encoded URLSegments were being truncated
* Encodes URLSegments before looking up Members by them

Fixes #465

### DIFF
--- a/code/model/Blog.php
+++ b/code/model/Blog.php
@@ -708,8 +708,10 @@ class Blog_Controller extends Page_Controller
         $urlSegment = $this->request->param('URLSegment');
 
         if ($urlSegment) {
+            $filter = URLSegmentFilter::create();
+
             return Member::get()
-                ->filter('URLSegment', $urlSegment)
+                ->filter('URLSegment', $filter->filter($urlSegment))
                 ->first();
         }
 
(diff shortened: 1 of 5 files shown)


code/model/Blog.php
```diff
@@ -709,7 +709,7 @@ public function getCurrentProfile()
 
         if ($urlSegment) {
             return Member::get()
-                ->filter('URLSegment', $urlSegment)
+                ->filter('URLSegment', rawurlencode($urlSegment))
```
micmania1: I think this should u se URLSegment filter. We fixed this previously for blog tag/categories.

Code: https://github.com/silverstripe/silverstripe-blog/blob/master/src/Model/BlogObject.php#L181
Issue: https://github.com/silverstripe/silverstripe-blog/issues/376 
PR: https://github.com/silverstripe/silverstripe-blog/pull/402

robbieaverill: Makes sense. I saw a couple f other places where this was done so followed, I will update them all to use the filter

Misagh90:
@robbieaverill I changed my blog.php & BlogMemberExtension.php
It's working
Thank you

PR #469, opened 2017-09-14
ENHANCEMENT -  SS4 compatibility 

fspringveldt:
This PR seeks to work on the SS4 support for this module. 

* updated deprecated Member::currentUser() references
* PHP ran
* removed @package and @subpackage references 
* Adds namespaces to _t() calls
* refactored to remove code from _config.php
* updated traditional array syntax
* updated deprecated Member::logout() calls
* replaced magic config() calls to use the config::get()
* refactored deprecated DB::getConn()
* refactored classes to use Injectable::create()

Doc updates coming in separate PR

### DIFF
--- a/src/Forms/GridField/GridFieldAddByDBField.php
+++ b/src/Forms/GridField/GridFieldAddByDBField.php
@@ -2,6 +2,7 @@
 
 namespace SilverStripe\Blog\Forms\GridField;
 
+use SilverStripe\Core\Manifest\ModuleLoader;
 use UnexpectedValueException;
 use SilverStripe\Control\Controller;
 use SilverStripe\Core\Convert;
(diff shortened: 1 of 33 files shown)


src/Forms/GridField/GridFieldAddByDBField.php
```diff
@@ -203,12 +201,12 @@ public function getHTMLFragments($gridField)
         $addAction->setAttribute('data-icon', 'add');
         $addAction->addExtraClass('btn btn-primary');
 
-        $forTemplate = new ArrayData(array());
+        $forTemplate = ArrayData::create([]);
```
robbieaverill: FYI we can omit the empty array argument now (just a comment)

tests/BlogCategoryTest.php
```diff
@@ -48,10 +49,10 @@ public function tearDown()
      */
     public function testBlogPosts()
     {
-        $member = Member::currentUser();
+        $member = Security::getCurrentUser();
 
         if ($member) {
-            $member->logout();
+            Security::setCurrentUser(null);
```
robbieaverill: This logic block could just be `$this->logOut();`. It's repeated a couple of times too.

PR #472, opened 2017-09-18
FIX updated translations to use namespaces

fspringveldt:


lang/en.yml
```diff
@@ -1,8 +1,15 @@
...
-    SINGULARNAME: 'Archive Widget'
-  Blog:
+  SilverStripe\Blog\Forms\GridField\GridFieldAddByDBField:
+    Add: 'Add {name}'
+    AddFail: 'Unable to save {class} to the database.'
+    PermissionFail: 'You don''t have permission to create a {class}.'
+  SilverStripe\Blog\Forms\GridField\GridFieldBlogPostState:
+    Draft: '<i class="btn-icon gridfield-icon btn-icon-pencil"></i> Saved as Draft on {date}'
+    Published: '<i class="btn-icon gridfield-icon btn-icon-accept"></i> Published on {date}'
+    Timer: '<i class="gridfield-icon blog-icon-timer"></i> Publish at {date}'
```
robbieaverill: This is unrelated, but it's a good time right now to remove the HTML from these translations

fspringveldt: Let me get on that.

PR #479, opened 2017-09-22
FIX Blog user settings help links displayed incorrectly

### LINKED ISSUE #474: SS4beta3 with Blog : raw html showing & broken title
The help links are showing raw html. Also, "Contributors" title is broken up at an odd place :)
screenshot from 2017-09-19 17-53-14

fspringveldt:
Fixes #474 

* Removes the raw HTML and restores the toggle functionality.
* Makes help text tranlatable
* Minimal styling changes
image

### DIFF
--- a/css/cms.css
+++ b/css/cms.css
@@ -74,16 +74,18 @@
   background: url("../images/information.png") no-repeat center center;
   width: 20px;
   height: 20px;
-  margin-left: 4px; }
+  margin-left: 4px;
+  cursor:pointer; }
 
 .middleColumn.toggle-description-correct-middle {
   margin-left: 0;
   float: left;
   width: 416px; }
 
-label.right.toggle-description-correct-right {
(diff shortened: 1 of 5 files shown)


css/cms.css
```diff
@@ -74,16 +74,18 @@
...
 .middleColumn.toggle-description-correct-middle {
   margin-left: 0;
   float: left;
   width: 416px; }
 
-label.right.toggle-description-correct-right {
+.tab-content .field p.toggle-description-correct-right {
   display: inline-block;
   margin-left: 0;
+  padding-left: 0;
```
robbieaverill: This is a dist file, you'll need to update the `blog/blog.scss` file and build it with Compass

dhensby: Hmm - there's no `config.rb` which makes this a bit difficult.

I've pushed a commit to fix this and compile

PR #480, opened 2017-09-22
FIX Display individual years in blog archive widget when set to "Yearly"

### LINKED ISSUE #366: Blog Archive Widget Has MySQLcentric Date Logic
The `DATE_FORMAT` function in the section following https://github.com/silverstripe/silverstripe-blog/blob/master/code/widgets/BlogArchiveWidget.php#L93 is MySQL only.

### LINKED ISSUE #464: Error with yearly archive in BlogArchiveWidget.php (2.4 Branch)
https://github.com/silverstripe/silverstripe-blog/blob/2.4/code/widgets/BlogArchiveWidget.php

related:
#453 
#446

Line 109:
`$date->setValue(strtotime($next['PublishDate']));`
when achive type is set to yearly `$next['PublishDate']` will be 2015, 2016, 2017 and so on
but strtotime won't return a correct time with only "YYYY" given
so `$year = $date->Format('Y');` in the next line will always return 2017, see screenshot
(monthly archive will work as then PublishDate is something like 2016-11)

29568061-72d86fb2-874f-11e7-91b0-94d45132e152

may this be a simple fix around line 112?
```
if ($this->ArchiveType == 'Yearly') {
                $year = $next['PublishDate']; //  #fix
                $month = null;
                $title = $year;
} else {
                $month = $date->Format('m');
                ...
                ...
``` #

robbieaverill:
Fixes #464 and fixes #366

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,7 @@ before_install:
 
 env:
   global:
-    - DB=MYSQL CORE_RELEASE=3.1
+    - DB=MYSQL CORE_RELEASE=3.6
 
     # Turn coverage off by default, as it's expensive time wise
     - COVERAGE=0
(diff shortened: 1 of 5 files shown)


robbieaverill:
I'll add a test for this, will be useful for the SS4 version which I don't believe has had this logic updated yet

robbieaverill:
Ok, I need to now fix this for SS 3.1

robbieaverill:
No, that's silly. I'm going to increase the minimum to 3.2.

PR #486, opened 2017-11-01
Allow a user to disable all profiles

zanderwar:
This module ships with User Profiles enabled by default.

By popular opinion (as per slack chat) user profiles aren't always a desired feature, this allows for them to be disabled _easily_.

Albeit imho, profiles should be an extension/module in `suggests` instead of factory standard

### DIFF
--- a/docs/en/configuring-user-profiles.md
+++ b/docs/en/configuring-user-profiles.md
@@ -1,0 +1,12 @@
+## Configuring user profiles
+
+This module ships with User Profiles enabled by default.
+
+If you'd prefer to disable this functionality and instead return a 404 for the `/profile/` page, you can do so using SilverStripe config.
+
+In mysite/_config/settings.yml
+
+```yaml
+SilverStripe\Blog\Model\BlogController:
+  disable_profiles: true
+```
--- a/src/Model/BlogController.php
+++ b/src/Model/BlogController.php
@@ -4,6 +4,7 @@ namespace SilverStripe\Blog\Model;
 
 use PageController;
 use SilverStripe\Control\Director;
+use SilverStripe\Control\HTTPResponse_Exception;
 use SilverStripe\Control\RSS\RSSFeed;
 use SilverStripe\ORM\ArrayList;
 use SilverStripe\ORM\DataList;
@@ -44,6 +45,14 @@ class BlogController extends PageController
     ];
 
     /**
+     * If enabled, blog author profiles will be turned off for this site
+     *
+     * @config
+     * @var bool
+     */
+    private static $disable_profiles = false;
+
+    /**
      * The current Blog Post DataList query.
      *
      * @var DataList
@@ -68,10 +77,16 @@ class BlogController extends PageController
     /**
      * Renders a Blog Member's profile.
      *
+     * @throws HTTPResponse_Exception
+     *
      * @return string
      */
     public function profile()
     {
+        if ($this->config()->get('disable_profiles')) {
+            $this->httpError(404, 'Not Found');
+        }
+
         $profile = $this->getCurrentProfile();
 
         if (!$profile) {
--- a/src/Model/BlogController.php
+++ b/src/Model/BlogController.php
@@ -6,6 +6,7 @@ use PageController;
 use SilverStripe\Control\Director;
 use SilverStripe\Control\RSS\RSSFeed;
 use SilverStripe\ORM\ArrayList;
+use SilverStripe\ORM\DataList;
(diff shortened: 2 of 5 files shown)


tests/BlogTest.php
```diff
@@ -348,6 +349,23 @@ public function testFilteredCategoriesArchive()
...
+            $controller = BlogController::create();
+            $controller->setRequest(Controller::curr()->getRequest());
+            $controller->profile();
+
+            $this->fail('The "profile" action should throw a HTTPResponse_Exception when disable_profiles is enabled');
+        } catch (HTTPResponse_Exception $e) {
+            $this->assertEquals(404, $e->getResponse()->getStatusCode(), 'The response status code should be 404 Not Found');
+        }
+
+        Config::modify()->set(BlogController::class, 'disable_profiles', false);
```
dhensby: This isn't needed as all configs are nested/unnested for each test.

tests/BlogTest.php
```diff
@@ -348,6 +349,23 @@ public function testFilteredCategoriesArchive()
...
+    {
+        Config::modify()->set(BlogController::class, 'disable_profiles', true);
+
+        try {
+            $controller = BlogController::create();
+            $controller->setRequest(Controller::curr()->getRequest());
+            $controller->profile();
+
+            $this->fail('The "profile" action should throw a HTTPResponse_Exception when disable_profiles is enabled');
+        } catch (HTTPResponse_Exception $e) {
```
dhensby: Please use the `expectException` annotation https://phpunit.de/manual/current/en/appendixes.annotations.html#appendixes.annotations.expectedException

zanderwar: Ah ok, How do those annotations work for checking the status code? I was following core code on [this one](https://github.com/silverstripe/silverstripe-cms/blob/4.0/tests/php/Controllers/CMSMainTest.php#L441)

src/Model/BlogController.php
```diff
@@ -67,10 +68,14 @@ public function index()
     /**
      * Renders a Blog Member's profile.
      *
-     * @return HTTPResponse
+     * @return string
      */
     public function profile()
     {
+        if ($this->stat('disable_profiles')) {
```
robbieaverill: `->stat()` is deprecated, can we please switch to `->get()` instead?

src/Model/BlogController.php
```diff
@@ -67,10 +68,14 @@ public function index()
...
+            $this->httpError(404, 'Not Found');
```
robbieaverill: Would you mind adding this to the doc block e.g. `@throws HTTPResponse_Exception If profiles are disabled`

docs/en/configuring-user-profiles.md
```diff
@@ -0,0 +1,12 @@
...
+
+This module ships with User Profiles enabled by default.
+
+If you'd prefer to disable this functionality and instead return a 404 for the `/profile/` page, you can do so using SilverStripe config.
+
+In mysite/_config/settings.yml
+
+```yaml
+SilverStripe\Blog\Model\BlogController:
+  disable_profiles: true
```
robbieaverill: It'd be lovely if you could also add this to the class and mark it as a config property:

```php
# File: BlogController.php
/**
 * If enabled, blog author profiles will be turned off for this site
 *
 * @config
 * @var bool
 */
private static $disable_profiles = false;
```

zanderwar: Done

src/Model/BlogPost.php
```diff
@@ -713,6 +714,16 @@ protected function getStaticCredits()
     }
 
     /**
+     * Checks to see if User Profiles has been disabled via config
+     *
+     * @return bool
+     */
+    public function getProfilesDisabled()
+    {
+        return (Config::inst()->get(BlogController::class, 'disable_profiles'));
```
dhensby: Don't wrap returns in parenthesis, please

tests/BlogTest.php
dhensby: Ah, good point - you're trying to assert the status code not just that the http response error is thrown

src/Model/BlogPost.php
zanderwar: Ok, was probably a bad idea. Wanted to force bool

robbieaverill: If you want to do that you could cast as bool e.g. `return (bool) ...`

zanderwar: It's very much the "same same" either way

dhensby:
thanks :)

PR #487, opened 2017-11-02
FIX Ensure that draft blog posts are always viewable to users with view draft permission

robbieaverill:
This change allows logged in users of any type with "view draft content" permission to view draft blog posts that are to be published in the future, where previously they would be denied without edit access.

Fixes https://github.com/silverstripe/silverstripe-sharedraftcontent/issues/70

### DIFF
--- a/code/model/BlogPost.php
+++ b/code/model/BlogPost.php
@@ -461,7 +461,12 @@ class BlogPost extends Page
             return false;
         }
 
-        if($this->canEdit($member)) {
+        if ($this->canEdit($member)) {
+            return true;
+        }
+
+        // If on draft stage, user has permission to view draft, so show it
+        if (Versioned::current_stage() === 'Stage') {
             return true;
         }
 
@@ -469,7 +474,7 @@ class BlogPost extends Page
          * @var SS_Datetime $publishDate
          */
         $publishDate = $this->dbObject('PublishDate');
-        if(!$publishDate->exists()) {
+        if (!$publishDate->exists()) {
             return false;
         }
 
--- a/tests/BlogPostTest.php
+++ b/tests/BlogPostTest.php
@@ -27,44 +27,67 @@ class BlogPostTest extends SapphireTest
     /**
      * @dataProvider canViewProvider
      */
-    public function testCanView($date, $user, $page, $canView)
+    public function testCanView($date, $user, $page, $canView, $stage)
     {
         $userRecord = $this->objFromFixture('Member', $user);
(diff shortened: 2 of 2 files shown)


code/model/BlogPost.php
```diff
@@ -457,19 +457,20 @@ public function canView($member = null)
     {
         $member = $this->getMember($member);
 
-        if (!parent::canView($member)) {
-            return false;
+        $parentCanView = parent::canView($member);
+        if (is_bool($parentCanView)) {
```
tractorcow: SiteTree::canView() should always return a bool, this isn't a very safe fix I'm afraid. All of the subsequent code would never be run.

tractorcow: Unfortunately, the only way I see around implementing this fix is inlining SiteTree::canView() in BlogPost::canView(), rather than relying on parent::canView().

code/model/BlogPost.php
```diff
@@ -457,19 +457,20 @@ public function canView($member = null)
...
+        if ($this->canEdit($member)) {
             return true;
         }
 
         /**
          * @var SS_Datetime $publishDate
          */
         $publishDate = $this->dbObject('PublishDate');
-        if(!$publishDate->exists()) {
+        if (!$publishDate->exists()) {
```
tractorcow: Could we change this check so that it ONLY ran on stage === Live? That would be a lot simpler right?

tractorcow:
```php
$isLive = $this->getSourceQueryParam('Versioned.mode') === 'stage'
    && $this->getSourceQueryParam('Versioned.mode') === Versioned::LIVE;
if ($isLive && !$publishDate->exists()) {
	//
}
```

robbieaverill:
Discussed with @tractorcow - summary:

**Problem**: users that have permission to view draft content cannot view a draft blog post if it has a publish date in the future (bug)

The user's permission to view draft content is checked and users rejected without it via `if (!parent::canView($member)) { return false; }`, so after that we can just check if it's the draft stage and return true.

If it's the live stage we would then continue to check for the publish date as per normal.

camfindlay:
assume this will get forward ported to the 2 branch at some point? Any eta on a 2.5.0? (awaiting a feature in 2.x-dev in a stable version)

camfindlay:
all good looks like it's already been ported... how bout that release? ;)

robbieaverill:
@camfindlay I'll try and get 2.5.0 out on hackday this Friday

PR #494, opened 2017-12-15
bugfix - set $table_name for widgets

jsirish:
SilverStripe 4.0.1 throws warnings when the widget module is installed.

### DIFF
--- a/src/Widgets/BlogRecentPostsWidget.php
+++ b/src/Widgets/BlogRecentPostsWidget.php
@@ -48,6 +48,11 @@ class BlogRecentPostsWidget extends Widget
     ];
 
     /**
+     * @var string
+     */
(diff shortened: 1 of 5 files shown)


src/Widgets/BlogRecentPostsWidget.php
```diff
@@ -47,6 +47,11 @@ class BlogRecentPostsWidget extends Widget
         'Blog' => Blog::class,
     ];
 
+    /**
+     * @var string
+     */
+    private static $table_name = 'RecentPostsWidget';
```
robbieaverill: Can we please call this `BlogRecentPostsWidget`, which will help ease SS3 to SS4 data migration

jsirish: sure thing, just pushed an update

PR #496, opened 2018-01-04
fixed namespacing for widgets

### LINKED ISSUE #495: namespacing for widgets seems to be broken

korthjp17:
fixes #495

### DIFF
--- a/src/Widgets/BlogArchiveWidgetController.php
+++ b/src/Widgets/BlogArchiveWidgetController.php
@@ -6,7 +6,7 @@ if (!class_exists('\\SilverStripe\\Widgets\\Model\\Widget')) {
     return;
 }
 
-use SilverStripe\Widgets\Controllers\WidgetController;
+use SilverStripe\Widgets\Model\WidgetController;
 
 class BlogArchiveWidgetController extends WidgetController
 {
--- a/src/Widgets/BlogCategoriesWidgetController.php
+++ b/src/Widgets/BlogCategoriesWidgetController.php
@@ -6,7 +6,7 @@ if (!class_exists('\\SilverStripe\\Widgets\\Model\\Widget')) {
     return;
 }
 
-use SilverStripe\Widgets\Controllers\WidgetController;
+use SilverStripe\Widgets\Model\WidgetController;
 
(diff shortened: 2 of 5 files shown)


robbieaverill:
Thanks for the fix. For context, the namespace was changed here: https://github.com/silverstripe/silverstripe-widgets/pull/163

robbieaverill:
Looks like there's a failure in the unit tests (which could easily be a pre-existing issue). Would you be keen to try and fix it?

korthjp17:
@mak001 is this something you can help me with?

robbieaverill:
I've just had a quick look at the test failure, here's a couple of notes:

* The fixture that is failing to be loaded is being successfully loaded in [a different test](https://github.com/silverstripe/silverstripe-blog/blob/master/tests/Widgets/BlogArchiveWidgetTest.php#L40) from the draft stage
* The failure is when trying to load the fixtured widget from the live stage
* The failing test published the owner BlogPost recursively (which should publish the widgets too) before switching the stage

I think in https://github.com/silverstripe/silverstripe-widgets/pull/163 we implemented the ownership  API into the widgets module, which is probably causing this error. It looks as if the way widgets are associated to blog pages in this module needs to be updated, since it's more like a has_one from Widget to Blog/BlogPost rather than using the intermediary WidgetArea, which is where the ownership API is centered.

I think this will need a little more work to fix. For now this pull request is a good improvement, so I think we can merge it.

@dhensby do you agree?

dhensby:
@robbieaverill yep, if the namespaces are wrong, lets merge that and worry about the tests separately.

dhensby:
Though I'm not sure why widgetcontrollers are in the Model namespace ;)

robbieaverill:
They don't _have_ to be, but [this logic](https://github.com/silverstripe/silverstripe-widgets/blob/master/src/Model/Widget.php#L247-L252) indicates that it makes more sense for them to live there for now

dhensby:
@robbieaverill whoops - this caused a regression in the test suite which we didn't spot :(

dhensby:
OK - maybe not this PR explicitly, maybe an upstream change - it looks like we aren't publishing a widget in the test when we need to.. PR incoming

PR #505, opened 2018-01-24
 FIX Replace jQuery UI icons with admin font icons

raissanorth:
Other changes:

* automatically publish a blog author's profile image when saving their profile, since the profile itself (Member object) is not versioned, so can't recursively publish via `$owns` API
* remove unused timer.png icon image (all replaced with font icons)

Example:

image

cc @sachajudd @clarkepaul


clarkepaul:
Is it possible to swap out one of the icons? from the looks of it the icon with the clock is meant to represent a future publish date but its using an icon which we associate with reverting something back to a certain time. The icon which should be used is the clock without the back arrow.
image

Looks good otherwise :)

raissanorth:
No problem @clarkepaul. Thanks for your feedback! I have changed it to the clock icon:

image

PR #509, opened 2018-02-19
Fixing non int pagination variable server error SS3

3Dgoo:
If a user tries to paginate the blog using a value that is not an integer SilverStripe will throw a server error.

Example. Visiting `blog/?start=10.1` will cause the following server error:

```You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '10.1' at line 8```

This change casts the pagination variable to an int before using it.

Note, this fix is for the `2` branch of the blog for SilverStripe 3.

### DIFF
--- a/code/model/Blog.php
+++ b/code/model/Blog.php
@@ -1043,7 +1043,7 @@ class Blog_Controller extends Page_Controller
         $posts->setPageLength($pageSize);
 
         // Set current page
-        $start = $this->request->getVar($posts->getPaginationGetVar());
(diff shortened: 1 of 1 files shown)


robbieaverill:
Merge on green

NightJar:
Hey Bob, thanks for the PR with a nice concise explanation as to what & why! Change looks good, I'll keep an eye on Travis to merge when we can :)

3Dgoo:
Thanks Dylan. Happy to help.

PR #512, opened 2018-03-04
ENHANCEMENT added getMinutesToRead()

zanderwar:
This PR wishes to introduce a method `getMinutesToRead` which returns an `integer` of how long an article will take to read.

Based on wikipedias answer to "How fast can a human read" of 200wpm.

> Source https://en.wikipedia.org/wiki/Speed_reading

I needed this for a client, and had achieved it with an extension but might go nicely in core

```twig
<div>$MinutesToRead Minute<% if $MinutesToRead > 1 %>s<% end_if %> Read</div>
```


zanderwar:
Sure I can add it to the EntryMeta template if that's what you'd like. I just wasn't 100% sure if that would be desired, glad to hear you like it :)

I'll get that in shortly.

zanderwar:
Sure

src/Model/BlogPost.php
```diff
@@ -755,6 +755,20 @@ public function getDate()
...
+    
+    /**
+     * Provides a rough estimate of how long this post will take to read based on wikipedias answer to "How fast can a
+     * human read" of 200wpm. Source https://en.wikipedia.org/wiki/Speed_reading
+     *
+     * @return string
+     */
+    public function getMinutesToRead()
+    {
+        $wpm = 200;
```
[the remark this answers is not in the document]
andrewandante: If you have a website primarily aimed at children, or older people, for example, you might want to nudge it down. A config var might also be appropriate?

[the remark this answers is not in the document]
robbieaverill: Good point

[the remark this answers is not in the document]
NightJar: I thought about config vars, but it's hard to change them on the fly from a template ;p
e.g. `<small>Approximately $MinutesToRead(100)` minutes reading time</small>`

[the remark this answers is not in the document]
dhensby: If we want to pass an argument in then the method name should be `MinutesToRead()` and not `getMinutesToRead` :)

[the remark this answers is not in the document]
zanderwar: All sounds great, will get to it tomorrow :)

zanderwar:
All done @dhensby

src/Model/BlogPost.php
```diff
@@ -756,6 +764,33 @@ public function getDate()
...
+     * human read" of 200wpm. Source https://en.wikipedia.org/wiki/Speed_reading
+     *
+     * @param null|integer $wpm
+     *
+     * @return string
+     */
+    public function getMinutesToRead($wpm = null)
+    {
+        if ($wpm && !is_numeric($wpm)) {
+            throw new \InvalidArgumentException(sprintf("Expecting integer but got %s instead", gettype($wpm)));
```
NightJar: I feel if you're going to go this far, it should perhaps come after setting from the config too.

zanderwar: I agree

src/Model/BlogPost.php
```diff
@@ -756,6 +764,33 @@ public function getDate()
...
 
+    /**
+     * Provides a rough estimate of how long this post will take to read based on wikipedias answer to "How fast can a
+     * human read" of 200wpm. Source https://en.wikipedia.org/wiki/Speed_reading
+     *
+     * @param null|integer $wpm
+     *
+     * @return string
+     */
+    public function getMinutesToRead($wpm = null)
```
dhensby: We need to rename this to `MinutesToRead` if we want to pass an argument through

src/Model/BlogPost.php
```diff
@@ -756,6 +764,33 @@ public function getDate()
...
+    {
+        if (!$wpm) {
+            $wpm = $this->config()->get('minutes_to_read_wpm') ?: 200;
```
dhensby: This ternary statement is not needed.

I'd replace these 3 lines with: `$wpm = $wpm ?: $this->config()->get('minutes_to_read_wpm');`

src/Model/BlogPost.php
zanderwar: You're right, done

src/Model/BlogPost.php
zanderwar: Done

src/Model/BlogPost.php
NightJar: Interestingly, that then makes it unpossible to use as a summary field - aliasing it is trivial, the question is more around whether or not it should already be extant. What do you think?

i.e. `public function getMinutesToRead() { return $this->MinutesToRead(); }`

zanderwar: I thought there was some magic that would look for either variant?

NightJar: Yeah, it's confusing and a bit inconsistent (to the likes of SSViewer), but the summary fields checks explicitly `hasMethod("get${name}")` if there's no matching _field_.

zanderwar: I mean the use case for the parameter would be very minimal in comparison with the configuration variable, could remove the parameter and return the `get*`

dhensby: I think it being used in summary fields is an edge case and one where the dev can add the `get` alias if needed...

src/Model/BlogPost.php
```diff
@@ -756,6 +764,31 @@ public function getDate()
...
+     */
+    public function MinutesToRead($wpm = null)
+    {
+        $wpm = $wpm ?: $this->config()->get('minutes_to_read_wpm');
+
+        if (!is_numeric($wpm)) {
+            throw new \InvalidArgumentException(sprintf("Expecting integer but got %s instead", gettype($wpm)));
+        }
+
+        $wordCount = count(array_filter(explode(" ", strip_tags($this->Content))));
```
dhensby: Can we use `str_word_count`?

Also, $this->Content won't have short codes parsed and `strip_tags` (probably) won't remove them? Do we think this is important or can we live with it?

NightJar: I can live with it. The number is of course an approximation anyway. I think going through all the hoops to render the content first is an expense we don't need.

zanderwar: It would be more accurate to parse the shortcodes then strip_tags. Will get this done

NightJar: I guess it's entirely feasible for a shortcode to expand into a wall of text... just not very likely I should hope.

dhensby:
thanks for working with me on this @zanderwar !

PR #515, opened 2018-03-20
Fix bad date formatting

DrMartinGonzo:
`m` would return minute in hour as seen here : http://userguide.icu-project.org/formatparse/datetime


DrMartinGonzo:
There might be some others hidden somewhere, but I don't have time to check right now :/

NightJar:
Hmm, been a bit preemptive on the merge there, but it would be great to also have a test to cover this.
I hear that you're short on time now, so I've raised this at #516 for someone to pick up in the (hopefully near) future.

PR #489, opened 2017-11-14
Summary field looked out of place.

MichaelJJames:
Summary field wasn't formatted correctly 

gtfzo8k

This css will make it look like 

gtfzqkq

PR #525, opened 2018-04-05
New commit for PR - #489

MichaelJJames:
@robbieaverill


robbieaverill:
This is just the dist file - are we missing the source change a well?

MichaelJJames:
Updated

robbieaverill:
Ok so this change looks alright on desktop but looks a bit off on smaller devices:

image
image
image
image

Might I suggest adding this as padding on the toggle composite field instead?

```css
#Form_EditForm_Summary_Holder {
  padding: 20px;
}
```

image
image
image
image

robbieaverill:
I actually feel like this could be addressed in the CMS module instead. Regardless, feel free to reopen the PR if you'd like to revisit it again =) Thanks for contributing @MichaelJJames!

PR #536, opened 2018-06-07
Call the correct method ->isModifiedOnDraft()

### LINKED ISSUE #535: The "modified" label is not working
https://github.com/silverstripe/silverstripe-blog/blob/06dbb72432e0004223cf58476a5880cfbfc4149a/src/Forms/GridField/GridFieldBlogPostState.php#L25

This have to be changed to $record->isModifiedOnStage() to call that function and check if the blogpost is modified.

MarcioBarrientos:
Fixes #535


robbieaverill:
I've rebased this onto the 3.1 branch so we can release it in a currently active version. Will merge on green. Thanks for the pull request!

PR #540, opened 2018-06-19
FIX Make sure `setAllowMultibyte` is on when looking up by URLSegment

### LINKED ISSUE #539: Problem with Tags when containing "(" etc
Is this perhaps an oversight or why is in URLSegmentExtension.php line 43 setAllowMultibyte set to true while in Blog.php line 864 (version 2.5) it is missing.

https://github.com/silverstripe/silverstripe-blog/blob/2.5.1/code/model/Blog.php#L851-L871

This works as long as there are no Chars like "(" or ")" in the Tag itself.
As soon as they are there they stop being found by the query because it removes the "(" and the database only knows what rawurlencode replaces them with i.e. %28.

dhensby:
fixes #539 

This needs merging up

### DIFF
--- a/tests/Widgets/BlogArchiveWidgetTest.php
+++ b/tests/Widgets/BlogArchiveWidgetTest.php
@@ -53,7 +53,9 @@ class BlogArchiveWidgetTest extends SapphireTest
             array('Title' => 'August 2017'),
         ), $archive);
 
-        Versioned::reading_stage($original);
+        if ($original) {
+            Versioned::reading_stage($original);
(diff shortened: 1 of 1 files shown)


chromos33:
Should this not also be done in getCurrentTag() ?

dhensby:
> Should this not also be done in getCurrentTag() ?

D'oh

dhensby:
in fact the member extension doesn't even turn mutlibyte on when generating URLs in the member extension!

PR #541, opened 2018-06-20
FIX Make sure all URL filters allow multibyte

### LINKED ISSUE #539: Problem with Tags when containing "(" etc

dhensby:
Fixes #539 

follow up to #540 - which didn't fully resolve the issues

### DIFF
--- a/code/extensions/BlogMemberExtension.php
+++ b/code/extensions/BlogMemberExtension.php
@@ -57,6 +57,8 @@ class BlogMemberExtension extends DataExtension
     public function generateURLSegment()
     {
         $filter = URLSegmentFilter::create();
+        $filter->setAllowMultibyte(true);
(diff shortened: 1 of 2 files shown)


robbieaverill:
Merge on green =)

NightJar:
Green.

PR #542, opened 2018-06-21
BlogTag existing check on URLSegment

gherkins:
perform existing check for BlogTag on URLSegment rather than Title. 
Title is non unique in the Database, thus allows (ci) duplicates in legacy entries

### DIFF
--- a/code/model/BlogTag.php
+++ b/code/model/BlogTag.php
@@ -88,7 +88,7 @@ class BlogTag extends DataObject implements CategorisationObject
             // Check for duplicate tags
             $blog = $this->Blog();
             if($blog && $blog->exists()) {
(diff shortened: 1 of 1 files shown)


dhensby:
The problem here, I believe, is that the interface has no way to determine the difference between the tags other than by title. Making this change, and thus allowing duplicates by Title, means that CMS users will see multiple of the same tag as options and not know which is the one they wanted.

dhensby:
You can see that this issues (#271) was actually raised because of the problem I'm stating above.

dhensby:
There's also a specific test to ensure duplicates by title are not allowed. Therefore, I'm going to close this as #notabug

PR #544, opened 2018-06-26
Fix blog archive widget bug

3Dgoo:
Fixes blog archive widget bug to only get the archive months for the blog linked to in this widget.

What was happening before is if you had more than one blog, the blog archive widget would get every blog post from every blog, create the archive month or year list from those blog posts and display that date list. This would result in some archive months (or years) that did not apply to the relevant blog. Clicking on those links would display an empty list.

This pull request filters the blog posts used for the archive loop by the widget blog ID.

### DIFF
--- a/code/widgets/BlogArchiveWidget.php
+++ b/code/widgets/BlogArchiveWidget.php
@@ -101,7 +101,7 @@ class BlogArchiveWidget extends Widget
             ->addGroupBy($publishDate)
             ->addOrderBy('"PublishDate" DESC')
             ->addLeftJoin('SiteTree' . $suffix, '"SiteTree' . $suffix . '"."ID" = "BlogPost' . $suffix . '"."ID"')
-            ->addWhere(array('"PublishDate" <= ?' => SS_Datetime::now()->Format('Y-m-d H:i:s'), 'SiteTree' . $suffix . '.ParentID' => $this->BlogID));
+            ->addWhere(array('"PublishDate" <= ?' => SS_Datetime::now()->Format('Y-m-d H:i:s'), '"SiteTree' . $suffix . '"."ParentID"' => $this->BlogID));
 
         $posts = $query->execute();
         $result = new ArrayList();
--- a/code/widgets/BlogArchiveWidget.php
+++ b/code/widgets/BlogArchiveWidget.php
@@ -100,7 +100,8 @@ class BlogArchiveWidget extends Widget
         $query = SQLSelect::create($fields, '"BlogPost' . $suffix . '"')
             ->addGroupBy($publishDate)
             ->addOrderBy('"PublishDate" DESC')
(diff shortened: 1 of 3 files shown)


robbieaverill:
Thanks for the PR! This makes sense. This part of the code has had to be maintained quite a bit over the last year or so - would you mind updating or writing new tests for this change as well?

3Dgoo:
Sure thing. I'll try to look into that when I get some time.

3Dgoo:
I have updated the archive widget test to have multiple blogs and posts across both blogs. I have altered the tests to check that the archive widget only displays the relevant posts.

code/widgets/BlogArchiveWidget.php
```diff
@@ -100,7 +100,8 @@ public function getArchive()
         $query = SQLSelect::create($fields, '"BlogPost' . $suffix . '"')
             ->addGroupBy($publishDate)
             ->addOrderBy('"PublishDate" DESC')
-            ->addWhere(array('"PublishDate" <= ?' => SS_Datetime::now()->Format('Y-m-d H:i:s')));
+            ->addLeftJoin('SiteTree' . $suffix, '"SiteTree' . $suffix . '"."ID" = "BlogPost' . $suffix . '"."ID"')
+            ->addWhere(array('"PublishDate" <= ?' => SS_Datetime::now()->Format('Y-m-d H:i:s'), 'SiteTree' . $suffix . '.ParentID' => $this->BlogID));
```
dhensby: You need to ansi quote the identifiers:

```diff
-->addWhere(array('"PublishDate" <= ?' => SS_Datetime::now()->Format('Y-m-d H:i:s'), 'SiteTree' . $suffix . '.ParentID' => $this->BlogID));
+->addWhere(array('"PublishDate" <= ?' => SS_Datetime::now()->Format('Y-m-d H:i:s'), "\"SiteTree{$suffix}".\"ParentID\"" => $this->BlogID));
```

3Dgoo: Thanks for pointing that out Dan. I have pushed a change to fix this up as per your suggestion.

NightJar:
@dhensby Would you like to re-review?

NightJar:
OK, I guess not :>The feedback was addressed anyway.
I think it looks good, I'll dismiss and merge :)

albanyacademy:
Was this reverted? I don't see these changes, and now I'm experiencing the bug as described in the OP - multiple blogs means the archive widget is querying all blogposts, and there's to param for ParentID or $this->Blog outside of generating the link.

robbieaverill:
Hi @albanyacademy, the change was merged into a SilverStripe 3 release line, not SilverStripe 4. I'll see if I can cherry-pick it in

albanyacademy:
yessssssssssssss thank you!

PR #556, opened 2018-10-10
Fix PostedIn translation

jmfederico:
"Publicado el", as it was written, relates to dates.
"Publicado en" relates to belonging.

We want the second translation, as it is used for example when PostedIn a category.

### DIFF
--- a/lang/es.yml
+++ b/lang/es.yml
@@ -31,7 +31,7 @@ es:
     PERMISSION_MANAGE_USERS_HELP: 'Permitir  de asignación de Editores, Escritores o Contribuidores a los blogs'
     PLURALNAME: Blogs
     Posted: Publicado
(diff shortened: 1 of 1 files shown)


robbieaverill:
Thanks for contributing! We manage our non-English translations through [Transifex](https://www.transifex.com/silverstripe/silverstripe-blog). If you wouldn't mind, could you please head over there and adjust this translation through their website? The language files in this module are automatically overwritten with the Transifex versions whenever we do a release, so I can merge this now but it may be reverted at the next release.

The specific line you're changing in this pull request is in Transifex here: https://www.transifex.com/silverstripe/silverstripe-blog/translate/#es/master-v2/119177876

jmfederico:
Changed in Transifex

robbieaverill:
Gracias!

PR #559, opened 2018-11-13
Add db index on PublishDate

stojg:
It's is quite common to order by the publish date or do where operations on this column. In many cases, this causes full table scans and other DB CPU intensive work.

In my opinion this should be a default for this module.


stojg:
I am a fool, will fix the code so it looks like proper php

robbieaverill:
@stojg if you could avoid making branches on origin in future that'd be sweet, we need to go and clean them up from packagist manually

stojg:
Apologies, I did not know that. Will make sure I stop with that from now on.

robbieaverill:
No stress, cheers

Issue #562, opened 2018-11-23
Incorrect date format used when dates are shown in templates.


shaunjc:
Hi Team,

The latest SilverStripe version uses ISO format when formatting dates, however, this module still uses PHP `DateTime` format, which isn't directly compatible, at least where it comes to displaying the current Month in Archive titles.

The lines from the next two example should be adjusted, and `F` should be replaced with `MMMM`.
https://github.com/silverstripe/silverstripe-blog/blob/5a355f2cf05f3e7fb99c1b237640a44ac99e0f22/src/Model/BlogController.php#L400
https://github.com/silverstripe/silverstripe-blog/blob/5a355f2cf05f3e7fb99c1b237640a44ac99e0f22/templates/SilverStripe/Blog/Model/Layout/Blog.ss#L12

Example with correct usage:
https://github.com/silverstripe/silverstripe-blog/blob/5a355f2cf05f3e7fb99c1b237640a44ac99e0f22/src/Widgets/BlogArchiveWidget.php#L141-L142

NightJar:
Hi @shaunjc, thanks for reporting :)
It seems like you've found a good solution to this issue already, how would you feel about making a pull request with the fix you've mentioned?

shaunjc:
Hi @NightJar,
While I'm happy to provide solutions to any issue I find, I'm not confident enough with Git to make a pull request.
I've made changes to the copy I downloaded using composer, as well as reported the issue to prevent any issues with future updates, and that should be sufficient for me at this point in time.

NightJar:
@shaunjc would you like some guidance on how to achieve this?
It's not too scary I promise :)

No issue if not, I should be able to find time to address this some time soon.

NightJar:
We chatted on slack, PR at #563

PR #563, opened 2018-11-26
FIX: Update Blog archive date formats for SS4

shaunjc:
The date formats for the blog archives were incompatible with the changes made in SS4. These changes should be required for all similar versions.


shaunjc:
Closes #562

PR #570, opened 2019-01-19
FIX: Add Blog::getControllerName() so /category links work

madmatt:
The `Blog` model doesn't include the required `getControllerName()` method, so it defaults to using `PageController`, meaning that links to blog categories (e.g. www.example.com/blog/category/example-category) doesn't work. Also broken include links to profiles, tags etc. - basically any functionality on BlogController doesn't work.

### DIFF
--- a/src/Model/Blog.php
+++ b/src/Model/Blog.php
@@ -700,4 +700,9 @@ class Blog extends Page implements PermissionProvider
 
         return $group;
     }
+
+    public function getControllerName()
+    {
+        return BlogController::class;
+    }
 }


robbieaverill:
Which version of SS is this using? This sounds like we might have broken something fundamental in core!

madmatt:
Awkward! Turns out this is because I had a leftover `Page::getControllerName()` from when I first upgraded this site and thought namespacing `Page` and `PageController` to be consistent was a good idea.

I think we can probably close this...

Issue #488, opened 2017-11-03
Multibyte URLs have stopped working


robbieaverill:
Similar to #484, but in SS4. The tests for them are also failing.

https://travis-ci.org/silverstripe/silverstripe-blog/jobs/296592633

Appears to have started failing in the last 24 hours or so. Previously passing master builds are now failing when re-run: https://travis-ci.org/silverstripe/silverstripe-blog/jobs/287015877

However, this build from yesterday passed: https://travis-ci.org/silverstripe/silverstripe-blog/jobs/296438910

NightJar:
Cannot reproduce, [builds now pass](https://travis-ci.org/silverstripe/silverstripe-blog/builds/334537119).

Tests [are very simple](https://github.com/silverstripe/silverstripe-blog/blob/ef2f1165d74e169073b8819874433796e0edfecb/tests/BlogFunctionalTest.php) and I cannot see any obvious reason for them to fail.
Nor can I [see any changes](https://github.com/silverstripe/silverstripe-blog/compare/271fbfeae75a1dc88ad940a9e8ac9af5d09f5c93...ef2f1165d74e169073b8819874433796e0edfecb) that might rectify any issues to do with routing.

NightJar:
Appears something changed somewhere that I didn't see, because it passes now and the old builds in OP still fail:
https://travis-ci.org/silverstripe/silverstripe-blog/jobs/287015877
https://travis-ci.org/silverstripe/silverstripe-blog/jobs/296592633

robbieaverill:
Resurfaced again since https://github.com/silverstripe/silverstripe-cms/pull/2365#issuecomment-460487655 was merged

robbieaverill:
PR at https://github.com/silverstripe/silverstripe-blog/pull/571
 and https://github.com/silverstripe/silverstripe-cms/pull/2384

robbieaverill:
Re-fixed again

NightJar:
@Cheddam has some follow up suggestions. i.e. fixed in the case that the tests cover, but probably not in the other two use cases that have the same old functionality.
That is, `profile` has been addressed, but `tags` and `catagories` have not.

PR #571, opened 2019-02-05
FIX Multibyte URL encoding for blog profiles, and encoded params in functional tests

robbieaverill:
* Functional tests should encode multibyte URL params before passing them to `get()`, to replicate real world scenarios
* Blog profile controller action does not need to double encode the argument

Fixes https://github.com/silverstripe/silverstripe-blog/issues/488

Requires https://github.com/silverstripe/silverstripe-cms/pull/2384 to be merged and merged up for tests to pass across all versions.

### DIFF
--- a/src/Model/BlogController.php
+++ b/src/Model/BlogController.php
@@ -106,12 +106,16 @@ class BlogController extends PageController
     public function getCurrentProfile()
     {
         $urlSegment = $this->request->param('URLSegment');
-
         if ($urlSegment) {
             $filter = URLSegmentFilter::create();
+            // url encode unless it's multibyte (already pre-encoded in the database)
+            // see https://github.com/silverstripe/silverstripe-cms/pull/2384
+            if (!$filter->getAllowMultibyte()) {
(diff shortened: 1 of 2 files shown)


Cheddam:
@robbieaverill The original PR that fixed this handling (back in 2017) included changes to the handling of not only the profile URL, but also categories and tags - see [here](https://github.com/silverstripe/silverstripe-blog/blob/3.2/src/Model/BlogController.php#L261) and [here](https://github.com/silverstripe/silverstripe-blog/blob/3.2/src/Model/BlogController.php#L307). I don't think we have tests covering these methods, so it's hard to tell, but I believe the adjustment to handling you made in this PR needs to be applied to both of the linked methods too.

robbieaverill:
Yeah true, there are tests for it. I'll update them now

robbieaverill:
So multibyte URLs were always working in categories and tags, regardless of the config. I've changed it to respect the config setting here: https://github.com/silverstripe/silverstripe-blog/pull/573

PR #573, opened 2019-02-07
FIX Categories and tags now respect multibyte url configuration

robbieaverill:
Previously multibyte URLs would always work, regardless of URLSegmentFilter::$default_allow_multibyte. This change means that they will only work when the setting is enabled.

Follow up from https://github.com/silverstripe/silverstripe-blog/pull/571

### DIFF
--- a/src/Model/Blog.php
+++ b/src/Model/Blog.php
@@ -17,6 +17,8 @@ use SilverStripe\Forms\NumericField;
 use SilverStripe\ORM\DataList;
 use SilverStripe\ORM\DataObject;
 use SilverStripe\ORM\DB;
+use SilverStripe\ORM\HasManyList;
+use SilverStripe\ORM\ManyManyList;
 use SilverStripe\ORM\SS_List;
 use SilverStripe\ORM\UnsavedRelationList;
 use SilverStripe\Security\Group;
--- a/src/Model/BlogController.php
+++ b/src/Model/BlogController.php
@@ -267,9 +267,14 @@ class BlogController extends PageController
         $tag = $this->request->param('Tag');
         if ($tag) {
             $filter = URLSegmentFilter::create();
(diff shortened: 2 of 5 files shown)


ScopeyNZ:
So blog always allows multibyte in essence? Maybe we should think about making this more in-line with CMS?

NightJar:
My perspective is that the filtering was done on the wrong side of the request.
That is, a request came in (i.e. a legit URL), and then is filtered (meaning it could end up different to what was requested), then searched for in order to fulfill the action.
I would presume the reason this was done would be to transliterate the URI - but I think that is incorrect behaviour in a URL. You either request the resource, or you don't. Don't effectively make it available on two URLs (one multibyte, one transliterated).

ScopeyNZ:
Ok after a quick chat we're wondering whether we even need to encode at all now? It makes sense to remove the existing encoding but is it even worth encoding at all if `getAllowMultibyte() === true`? It's unsupported anyway - there's not really any point. That should align with what CMS is doing I think?

robbieaverill:
@ScopeyNZ what's in this PR now does align with what the CMS is doing - in that it's not doing any extra encoding if getAllowMultibyte() is true.

ScopeyNZ:
Sorry this sentence:

> It makes sense to remove the existing encoding but is it even worth encoding at all if `getAllowMultibyte() === true`?

Should've read

> [...] if `getAllowMultibyte() !== true`?

It's probably something I'm misunderstanding. Feels weird to still have `rawurlencode` when the "support stuff that needs encoding" is off. Basically any multibyte URL won't match anyway so no need to encode.

Not a blocker if I don't know what's going on.

robbieaverill:
Yeah it’s a straight copy and paste from the core PR. I do think it’s a little odd that it’s not using the filter at all, but didn’t really think too far into it

PR #575, opened 2019-02-12
NEW Add tests for blog archive links and tidy up code re-use a little

### LINKED ISSUE #516: Increase test coverage to cover archive loading
via a functional test.
This should catch issues caused by e.g. Month 39 being requested via the URL as per the fix in #515

robbieaverill:
Resolves #516

### DIFF
--- a/tests/BlogPostTest.php
+++ b/tests/BlogPostTest.php
@@ -146,4 +146,42 @@ class BlogPostTest extends SapphireTest
         $this->expectException(\InvalidArgumentException::class);
         $blogPost->MinutesToRead('not-a-number');
     }
+
+    /**
+     * @param string $type
+     * @param string $expected
+     * @dataProvider monthlyArchiveLinkProvider
+     * @group wip
+     */
+    public function testGetMonthlyArchiveLink($type, $expected)
+    {
+        /** @var BlogPost $blogPost */
+        $blogPost = $this->objFromFixture(BlogPost::class, 'FirstBlogPost');
+
+        $archiveLink = $blogPost->getMonthlyArchiveLink($type);
+        $this->assertContains('archive/', $archiveLink);
+        $this->assertContains($expected, $archiveLink);
+    }
+
+    /**
+     * @return array[]
+     */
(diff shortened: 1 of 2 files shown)


tests/BlogPostTest.php
```diff
@@ -146,4 +146,42 @@ public function testMinutesToRead()
...
+    }
+
+    public function testGetYearlyArchiveLink()
+    {
+        /** @var BlogPost $blogPost */
+        $blogPost = $this->objFromFixture(BlogPost::class, 'FirstBlogPost');
+
+        $archiveLink = $blogPost->getYearlyArchiveLink();
+        $this->assertContains('archive/', $archiveLink);
+        $this->assertContains('/2013', $archiveLink);
```
NightJar: Should this not be a little stricter? Is there an `assertsEndsWith`?
All forms of the date in the URL will contain `/2013`

robbieaverill: I was trying to avoid being too strict, since modules like Fluent might modify URL structures

NightJar: 🤔 a fair point I hadn't considered.
But ... oh. Blacklisting extensions isn't really ~~possible~~ _feasible__, is it?
Is it possible to 'remove' all the extensions from a particular object instance before testing?

robbieaverill: Yeah: `$illegal_extensions = [Foo::class => '*']`. Personally I don't mind that these tests are forgiving though. This particular test is asserting that everything explicit that the method is doing gets done as we expect it to

This is 100% a unit test. I noted earlier that we can't easily make functional tests for this

NightJar: My only concern is that it could be possible to have `/2013/10/1` returned, and this test would still pass.
If that's OK then that's fine... out of context though I'm expecting it to be incorrect. If that's not the case, then perhaps a cheeky comment to specify that "extra date context is ignored" or something would be cool :)

robbieaverill: I've updated to use end of string assertions instead

PR #577, opened 2019-04-10
MINOR: avoid current blog post being included in the recent posts widget

spekulatius:
Hello @robbieaverill and @NightJar 

I've noticed the current blog post is included in the widget for the recent blog posts (if it is falling in the criteria for the most recent posts). I can't see a use case for linking back to the article you are currently viewing. The PR would fix this by excluding the current post from the list.

Cheers,
Peter


spekulatius:
Awesome, thanks for the quick review & merge @robbieaverill!

PR #585, opened 2019-05-13
BlogArchiveWidget limit fix

albanyacademy:
Add limit to the return $result. If NumberToDisplay is unset it doesn't limit it.

@robbieaverill you asked if I could apply the temp fix at the template level, but i figured if i could make a PR then i may as well just put it in where it ought to go.

### DIFF
--- a/src/Widgets/BlogArchiveWidget.php
+++ b/src/Widgets/BlogArchiveWidget.php
@@ -16,6 +16,7 @@ use SilverStripe\ORM\Queries\SQLSelect;
 use SilverStripe\Versioned\Versioned;
 use SilverStripe\View\ArrayData;
 use SilverStripe\Widgets\Model\Widget;
+use SilverStripe\Core\Convert;
 
 if (!class_exists(Widget::class)) {
     return;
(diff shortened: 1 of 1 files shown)


src/Widgets/BlogArchiveWidget.php
```diff
@@ -150,6 +150,6 @@ public function getArchive()
 
         $this->extend('updateGetArchive', $result);
 
-        return $result;
+        return $result->limit($this->NumberToDisplay);
```
robbieaverill: Nice! Maybe it'd be better to put the limit before the extend point, so extensions can continue to alter it afterwards

robbieaverill:
I've noticed that you've created this pull request from one of the "core branches" rather than a branch specifically intended for this pull request. This can cause problems if we ask you to target this PR at a different branch. No action is needed at the moment, this is for your information.

In the future please create a new branch for each individual pull request. See our [guide about contributing PRs](https://docs.silverstripe.org/en/4/contributing/code/#step-by-step-from-forking-to-sending-the-pull-request)

src/Widgets/BlogArchiveWidget.php
albanyacademy: added. moved it into an if statement as well.

question about the widgets: i see the limit applied in different ways. Categories and Tags have it wrapped in an if and the limit value is converted raw2sql - the posts widgets don't have those. I can see if being a bit faster if the limit() doesn't have to run, but is the Convert bit important?

albanyacademy:
updated, and apologies! first PR

src/Widgets/BlogArchiveWidget.php
```diff
@@ -148,6 +153,10 @@ public function getArchive()
             ]));
         }
 
+        if ($this->NumberToDisplay) {
```
robbieaverill: This part is a good change, but it looks like the changes above are part of a separate commit which is already part of the codebase - can you confirm why it's there?

emteknetnz:
@albanyacademy We’re currently going through pull-requests that haven’t had much activity for a while to see what is still active and what now makes sense to be closed.

Are you going to be doing further work on this or would you prefer to now close this one?

emteknetnz:
It seems like there's going to be no further activity on this pull request, so we've closed it for now. Please open a new pull-request if you want to re-approach this work, or for anything else you think could be fixed or improved.

PR #587, opened 2019-05-17
UX expand custom summary if field has value

### LINKED ISSUE #566: "Add a Custom Summary" should be uncollapsed by default if it is filled
"Add a Custom Summary" is at the bottom of the form, and collapsed by default. That's fine if it is not used, but less fine if it _is_ used. It becomes especially problematic if a translation module such as Fluent is used, because then you'll happily translate the Title and Content, but forget about the Custom Summary. 

After all, it says "*Add* a Custom Summary", so there's no visual indication whatsoever that something is filled in there. 

My suggestion is to simply uncollapse it (onload) if it has any content.

muskie9:
resolves #566

This only expands the toggle field if there is a value present. There was discussion of having it expanded all the time which would be simple enough to update.

### DIFF
--- a/src/Model/BlogPost.php
+++ b/src/Model/BlogPost.php
@@ -258,6 +258,10 @@ class BlogPost extends Page
             $summaryHolder->setHeadingLevel(4);
             $summaryHolder->addExtraClass('custom-summary');
 
+            if ($this->Summary) {
+                $summaryHolder->setStartClosed(false);
+            }
+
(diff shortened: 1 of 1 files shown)


ScopeyNZ:
Thanks @muskie9. Looks good. Would you mind rebasing on the `3` branch for the next minor release?

muskie9:
Don't think that worked how I thought it would lol

muskie9:
I think I've got it now, updated the branch to `3` and rebased

PR #590, opened 2019-06-30
Feature/gridfield add button name

Rhym:


stevie-mayhew:
Looks like the build error is unrelated to this? LGTM

src/Forms/GridField/GridFieldAddByDBField.php
```diff
@@ -188,14 +208,20 @@ public function getHTMLFragments($gridField)
             ->setAttribute('placeholder', $obj->fieldLabel($dbField))
             ->addExtraClass('no-change-track');
 
+        if (!$this->buttonName) {
```
tractorcow: This should probably be in a getter `getButtonName` with a fallback ($this->buttonName is only assigned by setButtonName) instead of assigned as a side-effect of getting the value.

Should probably be button Title, not button name (since name implies html identifier for this button, not the user-visible title).

src/Forms/GridField/GridFieldAddByDBField.php
```diff
@@ -188,14 +208,20 @@ public function getHTMLFragments($gridField)
...
         $addAction = GridField_FormAction::create(
             $gridField,
             'add',
             _t(
                 __CLASS__ . '.Add',
-                'Add {name}',
+                '{name}',
                 'Add button text',
-                ['name' => $obj->i18n_singular_name()]
+                ['name' => $this->buttonName]
```
tractorcow: Shouldn't button name be instead of the _t() call? This is acting as the object name, ("add {objectname}") not button name ("{buttonname}").

.gitignore
```diff
@@ -1,4 +1,5 @@
 /vendor/*
+/.idea
```
robbieaverill: We don't usually add things like this, since they're IDE specific. You can add them to your global gitignore though.

src/Forms/GridField/GridFieldAddByDBField.php
```diff
@@ -142,6 +149,19 @@ public function getDataObjectField()
         return $this->dataObjectField;
     }
 
+    /**
+     * Set the button name
+     *
+     * @param $name
```
robbieaverill:
```suggestion
     * @param string $name
```

src/Forms/GridField/GridFieldAddByDBField.php
tractorcow: You're going to end up with `add Add (objectname)` as your button...

src/Forms/GridField/GridFieldAddByDBField.php
```diff
@@ -188,14 +208,20 @@ public function getHTMLFragments($gridField)
             ->setAttribute('placeholder', $obj->fieldLabel($dbField))
             ->addExtraClass('no-change-track');
 
+        if (!$this->buttonName) {
+            // provide a default button name, can be changed by calling {@link setButtonName()} on this component
+            $objectName = $obj->i18n_singular_name();
+            $this->buttonName = _t(__CLASS__ . '.ButtonName', 'Add {name}', array('name' => $objectName));
```
robbieaverill:
```suggestion
            $this->buttonName = _t(__CLASS__ . '.ButtonName', 'Add {name}', ['name' => $objectName]);
```

src/Forms/GridField/GridFieldAddByDBField.php
```diff
@@ -188,14 +208,20 @@ public function getHTMLFragments($gridField)
...
+        }
+
         $addAction = GridField_FormAction::create(
             $gridField,
             'add',
             _t(
                 __CLASS__ . '.Add',
-                'Add {name}',
+                '{name}',
```
tractorcow: Don't change localisation default text. This will invalidate all the prior localised strings for this button.

Instead make the `_t()` call triggered in the fallback only, but the result becomes the third argument to `GridField_FormAction::create` directly.

src/Forms/GridField/GridFieldAddByDBField.php
tractorcow: Apparently GridFieldAddNewButton also calls this `ButtonName` so disregard that. :)

Rhym:
Bump

ScopeyNZ:
I just approved but now I'm wondering if `buttonName` is actually a good name for this attribute? It's more of a "model name override" or something? You can't actually alter the "Add " part.

Rhym:
> I just approved but now I'm wondering if `buttonName` is actually a good name for this attribute? It's more of a "model name override" or something? You can't actually alter the "Add " part.

@ScopeyNZ I just used the current naming conventions from https://github.com/silverstripe/silverstripe-blog/pull/590#issuecomment-509042200

ScopeyNZ:
Ah right I didn't connect to dots to the similar functionality. Perhaps it makes sense to move the whole `_t` bit to be the default value, meaning that if you set the button name you can control everything (including the `add` bit)? This would align with the functionality offered by `GridFieldAddNewButton` too.

NightJar:
Approved, so merged.
Any further followups can be added to `master`.

Issue #593, opened 2019-10-14
BUG New tags for blog post never get assigned to parent blog, result in orphaned duplicates


tractorcow:
Use case: I have a blog post A, in a Blog X

If I edit that posts and add a new tag, it gets created in the DB, but with a BlogID = 0.

If I now navigate to another blog post B in Blog X and edit tags, and search for that new tag, it never appears in the list. I create the same tag.

As a result, I end up with two tags with duplicate content in the DB, both with BlogID = 0.

Given that you can determine the tags in a blog by joining on blog posts with the ParentID of that blog, it seems hugely unnecessary to even require that tags be bound to a single (or orphaned) blog. It also means you have a huge number of duplicates (a version for a tag for each blog, plus many orphaned duplicates of that tag).

When assigning tags to a post, you also have no visibility about which tags are "safely" assigned to the parent blog, or whether they simply global tags.

My recommendation is this:
 - Drop has_one from both BlogTag and BlogCategory to Blog.
 - Re-implement the `Tags` and `Categories` methods on blog post to join against child posts instead.
- Blog Tags / Categories gridfields become read-only views.
- Add build task for resolving / merging duplicate tags and categories

tractorcow:
WIP at https://github.com/silverstripe/silverstripe-blog/pull/594

tractorcow:
I'm writing up the dedupe task now.

tractorcow:
https://github.com/silverstripe/silverstripe-blog/pull/594 is fixed and working.

robbieaverill:
Nice work @tractorcow, thanks again for your contribution

Issue #546, opened 2018-07-29
Tags and categories moved along post when moving post from one blog to another


clodeindustrie:
Sorry for the massive title.

When moving a post from one blog to another, the tags and categories of that post, seem to change ownership and become part of the destination blog.

They still stay attached to other post from the origin blog though.

raissanorth:
Thanks for reporting this @clodeindustrie. Could you please let us know which version of `silverstripe-blog` you are using, thanks.

raissanorth:
Hi @clodeindustrie,

thanks for your response.

I tried recreating this with 
```
silverstripe/blog                  3.0.0  
silverstripe/cms                   4.1.0
silverstripe/lumberjack            2.0.1
silverstripe/tagfield              2.0.0   
silverstripe/assets                1.1.0   
silverstripe/asset-admin           1.1.0  

```
but unfortunately could not reproduce. 

Steps taken:

1. I created two Blog sites, Blog 1 and Blog 2. 
2. I then created Blog Post A nested under Blog 1 and Blog Post B nested under Blog 2. 
3. Following that I assigned three categories and tags to Blog Post A. 
4. For the last step I moved Blog Post A to Blog 2 _and published it_. 

--> All of the categories and tags 'stayed' with Blog Post A. Blog Post B still had no categories nor tags associated with it. 

I have confirmed this in the CMS, frontend, and database. 

I'll close this for now, but if you continue to encounter this issue @clodeindustrie, please feel free to reopen (with another comment and steps to reproduce). Thanks!

PR #594, opened 2019-10-15
Category / Tag Rewrite - Remove dependency on Blog

tractorcow:
Initial implementation of #593 

This PR decouples tags and categories from blogs. When querying tags or categories for any blog, it will filter based on items that belong to current articles for that blog, rather than a distinct BlogID on each.

This prevents duplication of tags / categories (multiple blogs can share the same tag), and resolves a common data inconsistency where articles may be moved between blogs, but remain against tags / categories matching the "old" blog.

It also prevents issues with orphaned tags / categories (belonging to no blog at all).

Example inconsistency https://github.com/silverstripe/silverstripe-blog/issues/546

Next step is to build a task that de-duplicates tags / categories.

### DIFF
--- a/src/Admin/GridFieldCategorisationConfig.php
+++ b/src/Admin/GridFieldCategorisationConfig.php
@@ -4,6 +4,7 @@ namespace SilverStripe\Blog\Admin;
 
 use SilverStripe\Blog\Forms\GridField\GridFieldAddByDBField;
 use SilverStripe\Blog\Model\CategorisationObject;
+use SilverStripe\CMS\Model\SiteTree;
 use SilverStripe\Forms\GridField\GridFieldAddNewButton;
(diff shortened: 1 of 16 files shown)


tractorcow:
Pushed up a build task for cleaning up duplicates.

tractorcow:
I'm going to have to come back another day to fix these tests.

tractorcow:
Hurray it's green!

robbieaverill:
@Cheddam would you be keen to run your blog hiptest suite on this PR?

tractorcow:
> LGTM, what a lot of work! Nice one

I barely remember even writing this code...

I think I was very angry with Blog module that day and just got stuck in and fixed it. ;)

tractorcow:
@robbieaverill think we should merge this and make a new release? 

Not committer anymore so up to you bro lol.

I'm happy to use a composer fork for now though.

PR #597, opened 2019-11-03
Add missing widget models to db classname remapping

NightJar:
SilverStripe 4 has introduced namespacing, and thus ClassName field in
the database is now referring to classes that no longer exist. However
on a dev/build these will be updated so long as there is configuration
mapping the old class name to the new class name. However Blog module
here is missing the widget classes, which causes problems with
migrations. This will fix that.


robbieaverill:
Builds were already broken, merging

PR #608, opened 2020-04-16
Fix BlogPost canEdit permissions.

### LINKED ISSUE #607: Blog Post canEdit method allows some members to edit posts for which they are not an editor/author.
In `BlogPost.php`, the `canEdit` method immediately returns true if its parent's `canEdit` method returns true.
```php
if (parent::canEdit($member)) {
    return true;
}
```

This happens before any of BlogPost's bespoke checks (e.g. checking `$parent->isEditor`, `$this->isAuthor`, etc) which ultimately results in some members being able to edit _any_ blog post even if they're not an author for that post (depending on the permissions checks up the line).

It also means that if the parent says a member should _not_ be able to edit the page, if someone else has listed them as an author they may be able to edit the post anyway.

GuySartorelli:
resolves #607
If permissions earlier in the inheritance chain fail, we should not allow users to edit posts.
If permissions earlier in the inheritance chain succeed, we should still go through the checks in this method.

### DIFF
--- a/src/Model/BlogPost.php
+++ b/src/Model/BlogPost.php
@@ -564,8 +564,8 @@ class BlogPost extends Page
     {
         $member = $this->getMember($member);
 
-        if (parent::canEdit($member)) {
-            return true;
+        if (!parent::canEdit($member)) {
+            return false;
         }
 
         $parent = $this->Parent();


GuySartorelli:
It looks like the duplicate #792 has been merged in so I'll close this PR now.

GuySartorelli:
Had the wrong tab open, meant to close a different PR. Sorry.

dhensby:
The tests need updating to reflect the changes in logic

GuySartorelli:
Hey. Sorry, I've been trying to get phpunit working here but I can't figure out how to configure the database details for it, so it can't create the temporary database it needs to run tests.
Which is all just background information to say I can't fix the tests 'cause I can't run them.

GuySartorelli:
Hey there.
It is starting to look to me as though the tests which are failing _should_ be failing (e.g. BlogTest::testRoles, I think editor _should_ be able to edit that post. The canEdit check seems to be failing up the chain in SilverStripe's checking somewhere, and I can't make heads nor tails of why that would be.

Perhaps we should take a slightly different approach (which is already taken by the Blog class), and do our class specific tests first - then, only if we don't want to explicitly allow the user to edit the post (i.e. they're not listed as an author or editor for the post nor the parent blog), we can kick checks back to the superclass.

dhensby:
> It is starting to look to me as though the tests which are failing _should_ be failing (e.g. BlogTest::testRoles, I think editor _should_ be able to edit that post. The canEdit check seems to be failing up the chain in SilverStripe's checking somewhere, and I can't make heads nor tails of why that would be.

that's right, that's why the tests need updating, to reflect the new intended behaviour

GuySartorelli:
> > It is starting to look to me as though the tests which are failing _should_ be failing (e.g. BlogTest::testRoles, I think editor _should_ be able to edit that post. The canEdit check seems to be failing up the chain in SilverStripe's checking somewhere, and I can't make heads nor tails of why that would be.
> 
> that's right, that's why the tests need updating, to reflect the new intended behaviour

No, what I meant was that I think the tests as written are still correct, unless I have misunderstood the relationship between the test fixtures. Either I am misreading the fixtures, or my change has revealed a different issue. Either way I would appreciate someone taking a look and seeing what's going on.

emteknetnz:
@GuySartorelli are you still interested in this one? - note - it's targeting `master`

GuySartorelli:
Nope. It's something that should be fixed at some stage but I have no specific interest in diving into the tests to see what's going on there.
If someone else comes across this problem again and makes it known I might pick it up again but until then happy to close.

PR #621, opened 2020-08-25
FIX forward IDs given by DataList::relation()

### LINKED ISSUE #619: $this->Tags()->relation('BlogPosts') doesn't work
Currently I want to build a page exension that has a manymany relation to BlogTag so I can display related blog posts on other pages.

Unfortunately `$this->Tags()->relation('BlogPosts')` does not work, cause `DataList::relation()` tries to put some IDs when calling `BlogTag::BlogPosts()`, which is ignored by the BlogObject trait.

Fix in BlogObject: 

```
    /**
     * @return DataList
     */
    public function BlogPosts($id = null)
    {
        $blogPosts = parent::BlogPosts($id);

        $this->extend('updateGetBlogPosts', $blogPosts);

        return $blogPosts;
    }
```

## PRs:
  - [ ] https://github.com/silverstripe/silverstripe-blog/pull/621

wernerkrauss:
This way calling `->relation('BlogPosts')` does not return an UnsavedRelationList

fixes #619

### DIFF
--- a/src/Model/BlogObject.php
+++ b/src/Model/BlogObject.php
@@ -20,11 +20,12 @@ use SilverStripe\View\Parsers\URLSegmentFilter;
 trait BlogObject
 {
     /**
+     * @param int|array|null $id Optional ID(s) for parent of this relation, if not the current record
      * @return DataList
      */
-    public function BlogPosts()
+    public function BlogPosts($id = null)
     {
-        $blogPosts = parent::BlogPosts();
+        $blogPosts = parent::BlogPosts($id);
 
         $this->extend('updateGetBlogPosts', $blogPosts);
 


dnsl48:
Thank you for the PR!

## merge checklist
 
* [ ] the target branch is [correct](https://docs.silverstripe.org/en/4/contributing/code/#picking-the-right-version)
* [ ] all commits are relevant (e.g. no debugging and linting)
* [ ] the commit messages follow [the contribution guidelines](https://docs.silverstripe.org/en/4/contributing/code/#commit-messages)
* [ ] the patch follows [the contribution guidelines](https://docs.silverstripe.org/en/4/contributing/code/)
* [ ] new features are covered with tests (back-end with unit tests, front-end with Behat)
* [ ] the relevant CMS documentation is updated; changelog for impactful changes
* [ ] CI is green
* [ ] at least one peer reviewer approved; no changes requested

src/Model/BlogObject.php
```diff
@@ -21,11 +21,11 @@
 trait BlogObject
 {
     /**
-     * @return ManyManyList|BlogPost[]
+     * @return DataList
      */
-    public function BlogPosts()
+    public function BlogPosts($id = null)
```
dnsl48, line 26: Could you please add `@param int $id ...` to the method documentation?

dnsl48 (changes requested):
Looks good. Just need to add the new param documentation

dnsl48:
The branch `master` is reserved for a future MAJOR release, which is supposed to be version `4` of this module. However, I think this patch should end up in the next MINOR release, which is going to be `3.6`.

@wernerkrauss to do that, could you please retarget the branch `3` of this repository instead of `master`?

dnsl48:
@wernerkrauss to follow the commit message convention, could you add a prefix "FIX" to the commit message? So it may look like that: 
`FIX forward IDs given by DataList::relation()`

src/Model/BlogObject.php
wernerkrauss: I added the documentation and took the explanation from `DataObject::getComponents()`

src/Model/BlogObject.php
```diff
@@ -20,11 +20,12 @@
 trait BlogObject
 {
     /**
+     * @param int|array $id Optional ID(s) for parent of this relation
```
dnsl48, line 23:
```suggestion
     * @param int|array|null $id Optional ID(s) for parent of this relation
```

dnsl48 (changes requested):

dnsl48 (approved):
Thank you for the contribution!

Issue #614, opened 2020-07-03
Add Category and Add Tag button look weird


maxime-rainville:
Looks like these button and form field are meant to be fused together.

![image](https://user-images.githubusercontent.com/1168676/86432070-a57ec680-bd4a-11ea-8cc4-a2c98ceeb2b9.png)

PR #624, opened 2020-10-01
Fix category and tag buttons using input groups from bootstrap

MarcioBarrientos:
This fixes the category and tag button layout. #614 
Captura de Pantalla 2020-10-01 a la(s) 03 09 29

### DIFF
--- a/templates/SilverStripe/Blog/Forms/GridField/GridFieldAddByDBField.ss
+++ b/templates/SilverStripe/Blog/Forms/GridField/GridFieldAddByDBField.ss
@@ -1,5 +1,13 @@
 <div class="add-existing-autocompleter">
-	<% loop $Fields %>
-	<span>$Field</span>
-	<% end_loop %>
+    <div class="input-group">
+        <% loop $Fields %>
+            <% if $Type == 'action' %>
+                <div class="input-group-append">
+                    $Field
+                </div>
+            <% else %>
+                $Field
+            <% end_if %>
+        <% end_loop %>
+	</div>
 </div>


templates/SilverStripe/Blog/Forms/GridField/GridFieldAddByDBField.ss
```diff
@@ -1,5 +1,15 @@
...
-	<% loop $Fields %>
-	<span>$Field</span>
-	<% end_loop %>
+    <div class="input-group">
+        <% loop $Fields %>
+            <% if $Pos == 1 %>
+                $Field
+            <% end_if %>
+
+            <% if $Pos == 2 %>
```
dhensby, line 8: This doesn't seem like a good way to do this. Shouldn't it be more generic? Like if it's the last item or if it's a form action (button)?

dhensby (changes requested):
I think we need a more generic solution than relying on there being only 2 fields

templates/SilverStripe/Blog/Forms/GridField/GridFieldAddByDBField.ss
MarcioBarrientos: @dhensby I tried but I cant get a form action because it iterates an ArrayList and not a Form and always will be a 2 fields. Maybe this is better?
`<div class="add-existing-autocompleter">
    <div class="input-group">
        <% loop $Fields %>
            <% if not $Last %>
                $Field
            <% else %>
                <div class="input-group-append">
                    $Field
                </div>
            <% end_if %>
        <% end_loop %>
	</div>
</div>`

dhensby: You should be able to use `$Field.Type == 'action'`

MarcioBarrientos: @dhensby No, I can't get the Type property, I already did a debug and the only property that returns data is "Value", it returns a text input or the button without any indicator that I can use.

dhensby: Ah, have you tried just `$Type`?

MarcioBarrientos: @dhensby Yes, that did work

dhensby:
@MarcioBarrientos can you squash your two commits, please? Then I think this is good to merge

dhensby (approved):

PR #626, opened 2020-10-23
Using Advanced Workflows - If Blog post is in Workflow, Admin can still publish page breaking the workflow

obj63mc:
If you are utilizing the Advanced Workflows module using a simply review and approve process the following happens on all pages except blog posts currently.

1. Content author creates a 'Page' type page class and requests approval to publish
2. Admin is notified and must select 'reject' or 'approve', the publish button is hidden.  Screen shot below of how page normally looks for the admin -

Screen Shot 2020-10-23 at 11 07 19 AM

Now if we do the same with a blog post

1. Content Author creates a new 'Blog Post' type page class under a blog, requests approval to publish
2. Admin is notified and must select 'reject' or 'approve', the publish button is shown but should not be shown due to workflow being active.  Screen shot below -

Screen Shot 2020-10-23 at 11 09 19 AM

The cause for this is that the blog posts canPublish function was checking for an administrator account before letting the Advanced Workflow extensions run first.  I have changed the order and now this is working as expected.

### DIFF
--- a/src/Model/BlogPost.php
+++ b/src/Model/BlogPost.php
@@ -528,14 +528,14 @@ class BlogPost extends Page
     {
         $member = $this->getMember($member);
 
-        if (Permission::checkMember($member, 'ADMIN')) {
-            return true;
-        }
-
         $extended = $this->extendedCan('canPublish', $member);
 
         if ($extended !== null) {
             return $extended;
+        }
+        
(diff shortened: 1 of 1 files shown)


obj63mc:
Going to reopen this pull request but to the 3.5 branch as master seems to have a lot of different changes

PR #630, opened 2020-11-12
MNT Use shared travis config, use sminnee/phpunit

emteknetnz:


GuySartorelli:
I think this should be closed as it was superceded by #632 and other changes have also been made since.

PR #635, opened 2021-02-24
Ensure pagination start is not a negative number

3Dgoo:
If a user tries to access a negative start page such as `/blog?start=-10` this triggers a server error. This is because the code will pass this negative start to the limit function, which throws the error.

> [Emergency] Uncaught InvalidArgumentException: SQLSelect::setLimit() only takes positive values

This problem affects any Silverstripe website with a blog.

This code change should fix the issue.


emteknetnz (approved):
Looks good

PR #637, opened 2021-03-05
Only lazy load categories dropdown if there is a large number of categories

purplespider:
By lazy loading the dropdown, authors cannot see the list of existing categories to pick from and have to think of one from memory and start typing it.

This only sets the dropdown values to lazy load if there are more than 15 existing categories. 15 was chosen as this is also the number of categories that appear on one page of the Categories grid field.

### DIFF
--- a/src/Model/BlogPost.php
+++ b/src/Model/BlogPost.php
@@ -312,6 +312,11 @@ class BlogPost extends Page
                 );
             }
 
+            $shouldLazyLoadCategories = true;
+            if($this->Categories()->count() < 15) {
+                $shouldLazyLoadCategories = false;
+            }
+
             // Get categories and tags
             // @todo: Reimplement the sidebar
             // $options = BlogAdminSidebar::create(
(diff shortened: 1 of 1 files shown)


purplespider:
Is this good to be merged?

src/Model/BlogPost.php
```diff
@@ -312,6 +312,11 @@ public function getCMSFields()
                 );
             }
 
+            $shouldLazyLoadCategories = true;
+            if($this->Categories()->count() < 15) {
+                $shouldLazyLoadCategories = false;
+            }
```
dhensby, line 318: This can be changed to be a bit more readable (IMO) and also a bit better to phrase it in the positive sense "lazy load if more than 15" rather than currently "don't lazy load if less than 15".

```suggestion
            $shouldLazyLoadCategories = $this->Categories()->count() > 15;
```

dhensby (changes requested):
Stylistic change for you.

Also, why 15? That doesn't seem like a particularly large number for the application layer to handle/load. Even 100 should be reasonably safe to handle. What's the default pagination on a `GridField`, maybe that's a sensible number - alternatively, it could be configurable.

purplespider:
> Also, why 15?

Personally, I would have just set lazy loading to always be `false` for categories. I'd imagine that only in edge cases would there be so many as to cause an issue and it could be easily overridden in an extension then.

However, I assumed it was done for a reason, and as mentioned I just picked 15 as this is also the number of categories that appear on one page of the Categories grid field, which, I've just checked, [is actually](https://github.com/silverstripe/silverstripe-framework/blob/4795992208805aeb46723bb53b8dce88268af340/src/Forms/GridField/GridFieldPaginator.php#L27) the `$default_items_per_page` for `GridFieldPaginator`.

PR #639, opened 2021-03-25
PATCH: remove superfluous default value

sunnysideup:


emteknetnz:
Looks as though this default value is used by an [extension](https://github.com/silverstripe/silverstripe-comments/blob/3/src/Extensions/CommentsExtension.php#L97) in the comments module

I'll open a docs PR to explain this better

Issue #640, opened 2021-03-25
bug: blogpost is very slow when you have 5k Members


sunnysideup:
We had a site with 5k Members and the CMS kept running out of memory when we load a blog post.

michalkleiner:
This has been addressed in past IIRC, might be a regression that came back? There was a way how not to load all the members until interacting with the member dropdown. Unfortunately, I don't remember much more.

sunnysideup:
yes, I think you can change the set up in Blog to avoid this, but the thing is that this you have no idea why it loads so slow / trips over ... I'd say that the way it should be changed is that only members who can access the CMS can be authors ... I mean, this idea that any member can write a blog post is lovely, but feels more like and edge case (e.g. a members site where an author for a blog is just an ordinary members and somehow needs to be linked to blog that has been created and published by a CMS editor). 

here is my pull request: https://github.com/silverstripe/silverstripe-blog/pull/641

PR #641, opened 2021-03-25
MINOR: restrict the available authors, because right now any Member c…

sunnysideup:
…an be an author which causes big slowdowns for edge case only

see: https://github.com/silverstripe/silverstripe-blog/issues/640

### DIFF
--- a/src/Model/BlogPost.php
+++ b/src/Model/BlogPost.php
@@ -47,7 +47,7 @@ class BlogPost extends Page
      * given credit in the author field for blog posts
      * @var string|bool false or group code
      */
-    private static $restrict_authors_to_group = false;
+    private static $restrict_authors_to_group = 'CMS_ACCESS_CMSMain';
 
     /**
      * {@inheritDoc}


dhensby (commented):
This seems reasonable enough to me - how easy is it to revery this via the config if you *did* want all members shown? I think that should probably be added to the docs

dhensby:
You might argue that, if your site has lots of members, it's easier to make this tweak via config than it is to undo this tweak for a smaller site where you do want all users in the dropdown - though I agree that's quite an unlikely scenario

GuySartorelli:
My 2c: IMO this is a breaking change, if a minor one. Given that this is already a configuration option, it's probably best left as is. It can be set to a given group for any projects that need it already and if setting that one piece of configuration is becoming a real chore, you could always make a very small module which has Silverstripe/blog as a dependency and sets your opinionated default, and require that for your projects instead of the blog module directly.

dhensby:
Given this is targeting the master branch, any concerns about breaking changes aren't too important and this change should be considered on merit.

Personally, I think having authors that can't even access the CMS doesn't make much sense, if that's a specific use-case of a site, then having to tweak this to allow all members is pretty trivial.

I think my requests on this PR would be:

1. Tests
2. Documentation of the change and guidance how to bring back legacy behaviour

GuySartorelli:
Fair point, I hadn't noticed it was targeting master.

sunnysideup:
If Silverstripe is going to survive we need a more efficient way to add these changes.  This was raised in March last year. I knot that @GuySartorelli  and @dhensby are some of the few that actually are championing Silverstripe, but if you have a chance to raise this with your teams then please do so because this is a pretty trivial change and not adding it would add a lot of hassle to sites with many members and a blog ...

PR #651, opened 2021-04-29
Add index for `SiteTree` so that we prevent `index_merge` returning zero results

andrewandante:
This is a fix for a very strange issue that has been surfacing on Silverstripe Cloud platform.

It manifests as showing zero results when there are more than 3 `BlogPost`s on a given blog. 2 works just fine, as does 1, but three appears to be the magic number.

After a lot of weird debugging, I managed to narrow this down to MySQL using `index_merge` to retrieve the results of `BlogPost::get()->filter('ParentID', $this->ID)` - which returned inconsistent results. You can work around this by disabling `index_merge`, but that seemed like a bad idea, as it's the MySQL default.

Adding this index (which is the one that is ultimately generated by merging the two individual indexes) means that it doesn't have to create its own  "meta-index" and so the results are returned consistently.

I'm not _totally_ sure this works long-term (fix has only been in place a few days for each of the sites I've tested it on) but it has fixed multiple sites.

### DIFF
--- a/_config/indexes.yml
+++ b/_config/indexes.yml
@@ -1,0 +1,8 @@
+---
+Name: blogindexes
+---
+SilverStripe\CMS\Model\SiteTree:
+  indexes:
+    SiteTreeClassNameParentID:
+      - ClassName
+      - ParentID


jkersu:
Just putting my 2 cents that I have noticed the exact same issue on a site on SilverStripe Cloud which has been fixed as well by adding the index

michalkleiner:
Should this rather be done in silverstripe/cms repo?

andrewandante:
> Should this rather be done in silverstripe/cms repo?

Not sure to be honest - it comes up here most often because `BlogPost` looking for its parent `Blog` is such a prevalent use-case, and is directly related to the ORM call made on `init()` in `BlogController`.

There could be a wider idea to look at the most commonly-called queries and add some sensible default indexes - though I believe a lot of them already exist - but I think this makes the most sense as a helper for this specific call in this module. Happy to move it across if the overlords would prefer 😄

emteknetnz:
@andrewandante is this still a thing?

`MyDataObject::get()->filter('ParentID', $this->ID)` seems like an incredibly common query.  I we be extremely surprised to see this query failing if there are more than 3 `MyDataObject`s

I'd somewhat reluctant to add an index just to SiteTree given this only seem to affect BlogPost for whatever reason?

andrewandante:
I believe `->filter()` will expand out those args because of this: https://github.com/silverstripe/silverstripe-framework/blob/4/src/ORM/DataList.php#L395

I honestly don't know if this is still a thing - I suspect so? I don't really want to break a site that has been fixed just to test it. However, it also has clearly not affected every site, so maybe this is best preserved as a "this is how to fix this weird issue" issue, and close it. Up to you.

emteknetnz:
OK, I guess my understanding of `filter()` was wrong

I'll close this for now.  If people comment on this saying this is still a thing and provide further insight we can re-open and investigate

PR #654, opened 2021-05-13
Fix typo in custom-publish-dates.md

manuth:


dhensby:
failure is pre-existing

PR #655, opened 2021-05-13
Fix typo in using-widgets.md

manuth:


dhensby:
merging as master failures are pre-existing and this is just a readme update

PR #657, opened 2021-05-19
FIX BlogPost::onBeforePublish() should also trigger parent::onBeforePublish() if exists

christopherdarling:
e.g. if Page::onBeforePublish() exists.

### DIFF
--- a/src/Model/BlogPost.php
+++ b/src/Model/BlogPost.php
@@ -493,6 +493,10 @@ class BlogPost extends Page
      */
     public function onBeforePublish()
     {
+        if (method_exists(get_parent_class($this), 'onBeforePublish')) {
+            parent::onBeforePublish();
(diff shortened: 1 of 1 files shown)


dhensby:
This won't work if the parent class has the method added via an extension.

christopherdarling:
@dhensby  `Versioned::publishSingle()` calls `invokeWithExtensions()` so if `onBeforePublish()` is added via a `DataExtension` this should still be triggered, no?

dhensby:
Would it? It would see that it exists on `BlogPost` and so will invoke it there and that would be it... right? If you also added an extension to the parent or to `BlogPost` would that not be ignored?

andrewandante:
Going to close this as it's targeting CMS4, which is well out of support

PR #661, opened 2021-10-12
API phpunit9 support

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-framework/issues/10019

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -8,6 +8,7 @@
     ],
     "type": "silverstripe-vendormodule",
     "require": {
+        "php": "^7.3 || ^8.0",
         "silverstripe/cms": "^4.0",
         "silverstripe/lumberjack": "^2.0",
         "silverstripe/tagfield": "^2.0",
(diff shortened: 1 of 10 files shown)


composer.json
```diff
@@ -15,7 +15,7 @@
         "silverstripe/asset-admin": "^1.0"
     },
     "require-dev": {
-        "sminnee/phpunit": "^5.7",
+        "phpunit/phpunit": "^9",
```
maxime-rainville, line 18:
```suggestion
        "phpunit/phpunit": "^9.5",
```

maxime-rainville (changes requested):
Looks good, just need to make the PHPUnit constraint `^9.5`. Maybe add the PHP constraint while you're in there.

composer.json
emteknetnz: Updated

maxime-rainville (approved):
LGTM

Issue #484, opened 2017-10-17
Umlaut in BlogCategory/-Tag/Profile throws a 404


lerni:
getCurrentCategory, getCurrentTag, getCurrentProfile all use a URLSegmentFilter. 

URLSegmentExtension also uses a URLSegmentFilter in generateURLSegment and sets it to AllowMultibyte(true) but those above not. Hence a 404 it thrown say for example in:
https://github.com/silverstripe/silverstripe-blog/blob/2.4/code/model/Blog.php#L893

Wouldn't it make sense to $filter->setAllowMultibyte(true); for all of the above?

https://github.com/silverstripe/silverstripe-blog/blob/2.4/code/model/Blog.php#L711
https://github.com/silverstripe/silverstripe-blog/blob/2.4/code/model/Blog.php#L911
https://github.com/silverstripe/silverstripe-blog/blob/2.4/code/model/Blog.php#L911

Or how would you approach this? If my suggestion makes sense I'm happy to PR.

robbieaverill:
Hi @lerni, thanks for reporting.

Yeah - we've had a couple of patches go in recently for this kind of thing - perhaps haven't caught all of them yet!

The "allow multibyte" setting is configurable, and URLSegmentFilter applies it by default if it's enabled, so I don't think enabling it explicitly by default would be appropriate.

Are you experiencing this in an environment with "allow multibyte" enabled, or disabled?

If you're willing to submit a pull request, it'd be ideal if you could also provide a unit test with any code changes that exemplify the problem you're experiencing, and pass with your changes.

Cheers!

lerni:
ciao @robbieaverill
thx for answering! setting default_allow_multibyte to true per config resolves the issue with false it persists and a 404 is thrown for example on /blog/category/röbi
to be honest I won’t find the time to provide a unit test anytime soon, since I'm not familiar enough with testing. But if I get around one day the test needs to show the failure and the PR could set default_allow_multibyte: true in config.yml. Such a change could go into blog 2.5 which is probable not that likely to happen? I need it on SS 3.6.x. So it basically means if one need Umlaute in categories/tags or profile, you need to set it your self for 2.x?

robbieaverill:
@lerni 

>setting default_allow_multibyte to true per config resolves the issue with false it persists and a 404 is thrown for example on /blog/category/röbi

Ok, I think this is expected behaviour. It's up to the user to enable multibyte filtering.

[...]

PR #685, opened 2022-08-22
Allow multibyte to be configured (fixes #605)

### LINKED ISSUE #605: Special characters in Category/Tag URLSegment not working
When creating a Category/Tag for a blog through the CMS, URLSegmentFilter allows special characters (unlike the silverstripe default) and saves it as url encoded string.
However, when visiting the page, `getCurrentCategory()` searches for a non-urlencoded string and for a string that is filtered by `URLSegmentFilter` without allowing special characters.

As a result, creating category `küche` will save as `k%C3%BCche`. But when visiting `/blog/category/k%C3%BCche` then `$this->request->param('Category')` in `Blog->getCurrentCategory()` will actually return `küche` (non url encoded). And will search for the category with `->filter('URLSegment', ['küche', 'kueche'])`, which returns nothing.

I realise that there is a reason (#376) for allowing url encoded strings to be used. But currently that's not actually working.

We should either:
1. change how the query is handled when visiting the page (eg do `->filter('URLSegment', ['küche', 'kueche', 'k%C3%BCche'])`
2. change how the URLSegment is created and use URLSegmentFilter like the rest of the CMS (I'd prefer that way, but we need to come up with a way how to handle empty strings)

wilr:


src/Model/BlogObject.php
```diff
@@ -181,12 +181,7 @@ public function generateURLSegment($increment = 0)
     {
         $increment = (int) $increment;
         $filter = URLSegmentFilter::create();
-
-        // Setting this to on. Because of the UI flow, it would be quite a lot of work
-        // to support turning this off. (ie. the add by title flow would not work).
-        // If this becomes a problem we can approach it then.
-        // @see https://github.com/silverstripe/silverstripe-blog/issues/376
-        $filter->setAllowMultibyte(true);
+        $filter->setAllowMultibyte($this->config()->get('allow_multibyte') ?? true);
```
emteknetnz: Where is the `allow_multibyte` config defined?  I couldn't find it?  The closest config I could find was `default_allow_multibyte` which is defined on `URLSegmentFilter`.

If this is just a random, non-API piece of config then I'm not keen on this.

Also, if we do allow turning this off via config, based on the code comment there's further UI work that also needs to be done?

emteknetnz (changes requested):

src/Model/BlogObject.php
wilr: `BlogObject` is a trait which can be applied to anything Categories, Tags out of the box but anything project specific such as topic or (particular author profile sections in our case). So it doesn't make much sense for it to be defined somewhere.

As far as I can tell turning it off has no issues on our site but this change leaves the default as is so whatever issue there is can be looked at in the future.

emteknetnz: Are there any other examples of "undefined trait config" being used anywhere? If there are then this is probably OK, though if not I'm not keen to introduce a new way of using config, given peoples existing expectations that each piece of config has a corresponding `private static` that defines it.

wilr: `Are there any other examples of "undefined trait config" being used anywhere?`

Probably not yet because we don't use traits a huge amount at this stage in core.

emteknetnz: What happens if you put the `private static $allow_multibyte = true` on the trait?  Is the config not added to the classes the that use the trait - meaning you'd still need the `$this->config()->get('allow_multibyte') ?? true` style check?

Perhaps it would have been better if the trait was a DataExtension so that config is automatically applied to DataObjects it is applied to.  Too late for that I suppose.

I have a pretty big objection to undefined 'dynamic' API such as https://php.watch/versions/8.2/dynamic-properties-deprecated which has caused issues on the queued-jobs module.  As a maintainer who obsesses over semver and backwards compatibility, dynamic API is pretty awful.

Would it work if the trait had something like this so that it works with the existing (admittedly clunky) way of working?

**BlogObject.php**
```php
private function getAllowMultibyte(): bool
{
    $allowMultiByte = true;
    $this->extend('updateAllowMultibyte', $allowMultiByte);
    return $allowMultiByte;
}
```
And then add a extension classes to set this to false on Categories, Tags, etc?

wilr: Seems vastly over the top and complex for what is (hopefully) a short term work around within a patch release but if that's what works then happy to do it via extension hooks.

lerni: Movement on this would be nice. A dedicated URLSegment proved to work best for me. Not sure if it's still the case but also `RedirectorPage` used to break with global mulitibyte Links set. https://github.com/silverstripe/silverstripe-blog/issues/484#issuecomment-342806753

GuySartorelli:
I'm going to close this due to lack of action. If someone would like to pick up this work feel free to create a new PR with the requested change applied.

wilr:
Hi @GuySartorelli What diud you want to see changed? As per my last comment the requested changes I don't agree with and as @lerni pointed out - the fix does what it says on the tin

wilr:
> Perhaps it would have been better if the trait was a DataExtension so that config is automatically applied to DataObjects it is applied to. 

Perhaps it would be that would require a major API change compared to the current proposed fix which retains existing API.

GuySartorelli:
Sounded like Steve was leaning towards preferring an extension to update the value, e.g.
```php
$allowMultiByte = true;
$this->extend('updateAllowUrlSegmentMultibyte', $allowMultiByte);
$filter->setAllowMultibyte($allowMultiByte);
```

Personally I do prefer the config approach, so I'd probably be okay with your existing PR if the config name was a little more specific - `allow_multibyte` could easily be used as a config name for allow multibyte for just about any purpose entirely unrelated to this specific use case. If you changed it so something more like `allow_urlsegment_multibyte` I would be okay with it.

It _is_ preferable to declare `private static` for new config, but you can't override it on the class if you add that to the trait. Perhaps the solution that would satisfy Steve then is to add those `private static` declarations on the classes included in this module which _do_ use the trait, so it is explicitly declared where we have the control to do so.

wilr:
@GuySartorelli Done.

src/Model/BlogCategory.php
```diff
@@ -7,7 +7,7 @@
 /**
  * A blog category for generalising blog posts.
  *
-*
+ *
```
ChloeSartorelli:
```suggestion
```
May as well just remove that - no need to have two empty lines here.
I won't hold back merging if this isn't done though.

src/Model/BlogCategory.php
```diff
@@ -33,6 +33,13 @@ class BlogCategory extends DataObject implements CategorisationObject
      */
     private static $table_name = 'BlogCategory';
 
+    /**
+     * @config
+     *
+     * @var boolean
+     */
+    private static $allow_urlsegment_multibyte = true;
```
ChloeSartorelli:
```suggestion
    private static bool $allow_urlsegment_multibyte = true;
```
- `@config` is implied and is likely to be removed in a future major release anyway.
- We prefer to explicitly type-hint new API, even config. It's not enforced currently but could be in the future, and typehinting now will reduce the effort to have full typehinting across the board in the future.

ChloeSartorelli: Please do this for `BlogTag` as well.

GuySartorelli (changes requested):
Nice, thanks for that. Just a small change to make per below.

GuySartorelli (approved):
Looks good to me. Thanks.

lerni:
Now in BlogController.php `rawurlencode` is triggered based on the mb-config from `URLSegmentFilter` but it should respect `SilverStripe\Blog\Model\BlogTag` & `SilverStripe\Blog\Model\BlogCategory: allow_urlsegment_multibyte` config value in order to show BlogPosts associated with Tags/Categories that actually have mb-characters like umlauts in it.

PR #697, opened 2023-01-20
DEP Upgrade frontend build stack

GuySartorelli:
Relies on the following additional PRs to work: 
- https://github.com/silverstripe/silverstripe-behat-extension/pull/236
- https://github.com/silverstripe/silverstripe-widgets/pull/194
- https://github.com/silverstripe/silverstripe-tagfield/pull/229
- https://github.com/silverstripe/silverstripe-comments/pull/358 

## Parent issue 
- https://github.com/silverstripeltd/product-issues/issues/644


emteknetnz (changes requested):
Failing behat tests after rerun after merging other PRs

GuySartorelli:
I can reproduce the failures locally. Will look at them next week.

GuySartorelli:
This failure also happens when running behat tests without this PR (if you update the feature to correctly set the categories and tags). Whatever is causing the failure, it's not caused by this PR.
The behaviour the behat test is covering _does_ work when attempted manually and I can't quickly find what's causing the failure, so I think we should treat this in a separate broken builds issue.

emteknetnz (approved):

PR #700, opened 2023-01-31
FIX Validate Category / Tag Titles 

### LINKED ISSUE #695: The user can create and save a Blog Category or Blog Tag without a Title
### Description
When a user creates a blog, in the Categorisation tab he can add a Blog Category or Blog Tag without a title and then save empty values.

### AC's
- Validation that the Title field in Tag / Category is valid at the time the blog post is saved is added;
- Tests checking that these fields are not empty are added;

freezernick:
Closes #695

### DIFF
--- a/src/Model/BlogObject.php
+++ b/src/Model/BlogObject.php
@@ -72,7 +72,7 @@ trait BlogObject
             $validation->addError($this->getDuplicateError(), self::DUPLICATE_EXCEPTION);
         }
 
-        if(empty($this->Title)) {
+        if (empty($this->Title)) {
             $validation->addError($this->getEmptyTitleError(), self::EMPTY_TITLE_EXCEPTION);
         }
 
--- a/src/Model/BlogObject.php
+++ b/src/Model/BlogObject.php
@@ -72,6 +72,10 @@ trait BlogObject
             $validation->addError($this->getDuplicateError(), self::DUPLICATE_EXCEPTION);
         }
 
+        if(empty($this->Title)) {
+            $validation->addError($this->getEmptyTitleError(), self::EMPTY_TITLE_EXCEPTION);
+        }
+
         return $validation;
     }
 
@@ -244,4 +248,14 @@ trait BlogObject
      * @return string
      */
     abstract protected function getDuplicateError();
+
+    /**
+     * Returns an error message for this object when it tries to write with an empty title.
+     *
+     * @return string
+     */
+    protected function getEmptyTitleError()
+    {
+        return _t(__CLASS__ . '.EmptyTitle', 'Title must not be empty');
+    }
 }
--- a/src/Model/BlogCategory.php
+++ b/src/Model/BlogCategory.php
@@ -27,6 +27,8 @@ class BlogCategory extends DataObject implements CategorisationObject
      */
     const DUPLICATE_EXCEPTION = 'DUPLICATE';
 
+    const EMPTY_TITLE_EXCEPTION = 'EMPTY_TITLE';
+
     /**
      * {@inheritDoc}
      * @var string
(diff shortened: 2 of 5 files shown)


src/Model/BlogObject.php
```diff
@@ -72,6 +72,10 @@ public function validate()
             $validation->addError($this->getDuplicateError(), self::DUPLICATE_EXCEPTION);
         }
 
+        if(empty($this->Title)) {
```
michalkleiner:
```suggestion
        if (empty($this->Title)) {
```

GuySartorelli (commented):
@silverstripe/core-team 
What's our thoughts about this in regards to semver?
Specifically, changing the validation rules for whether an empty value is valid or not.

On one hand, it seems like there's no valid reason to have an empty category/tag - but on the other, someone could very well be going `BlogTag::create()->write()` in a unit test, if the _value_ of the tag doesn't matter for that test.

src/Model/BlogObject.php
```diff
@@ -72,6 +72,10 @@ public function validate()
             $validation->addError($this->getDuplicateError(), self::DUPLICATE_EXCEPTION);
         }
 
+        if (empty($this->Title)) {
+            $validation->addError($this->getEmptyTitleError(), self::EMPTY_TITLE_EXCEPTION);
```
ChloeSartorelli, line 76:
```suggestion
            $validation->addError($this->getEmptyTitleError(), self::EMPTY_TITLE_EXCEPTION ?? 'EMPTY_TITLE');
```
The constant won't be declared on any custom classes that use this trait, and requiring it to be added is technically a breaking change.

Note: Please test that this actually works - I've not tried the null coalescing operator in this context before. I think it'll work but it'll pay to check.

GuySartorelli (changes requested):

src/Model/BlogObject.php
emteknetnz: Yikes, shouldn't be calling `self::<const>` on a trait.  Not a great choice when this was made back in the day

I don't think null coalescing works here

```php
class C { function F(){ var_dump(self::Z ?? 'default'); } }; (new C)->F();
// PHP Warning:  Uncaught Error: Undefined constant C::Z in php shell cod
```

emteknetnz:
>> someone could very well be going BlogTag::create()->write() in a unit test

No one other than core team is going to write new unit tests, and the existing unit tests are green on this PR, so I see no reason to be concerned here.

GuySartorelli:
> No one other than core team is going to write new unit tests

People write unit tests for their own projects and modules all the time. It's not super likely those tests will rely on blog tags or categories but it's also not impossible.

emteknetnz:
Oh sorry I meant "No one other than core team is going to write new unit tests _for the blog module_" :-) 

What's the concern? That a project will create their own unit test that instantiates a their own subclass of a  `BlogCategory` or `BlogTag` with a blank Title?  That's seems extremely unlikely.  And even if they did and their test fails, they'll get a pretty clear `Title must not be empty` error which should be easy to trace back, so fixing the broken tests should require minimal effort.

While we're very strict with respecting semver API in minor releases, historically we've been somewhat lenient/pragmatic with behavioral changes, including changes to default configuration values.  This PR seems at the low end of changes within a minor.

michalkleiner:
> What's our thoughts about this in regards to semver? Specifically, changing the validation rules for whether an empty value is valid or not.
> 
> On one hand, it seems like there's no valid reason to have an empty category/tag - but on the other, someone could very well be going `BlogTag::create()->write()` in a unit test, if the _value_ of the tag doesn't matter for that test.

Not bothered by this particular change, I agree with Steve on this.

GuySartorelli:
Sweet, just wanted to get the general feeling about it, since I was on the fence about it.
There's problems with how the constant is being applied, but sounds like the general approach is okay.

GuySartorelli:
@freezernick Are you still interested in this pull request being merged? If so, please make changes to how the constant is applied as requested above.

GuySartorelli:
I'm closing this due to inactivity. If you want to work on it again I'll be happy to reopen it, or you can open a new pull request from scratch.

PR #720, opened 2023-08-10
Replace deprecated sort rand with shuffle

3Dgoo:


GuySartorelli (changes requested):
Hi, thanks for this.
It would have been good to include more information about the rand function being deprecated such as a link to a relevant changelog - but regardless, this seems like a sensible change.

Please retarget the pr to the 3.12 branch, as it seems like this is a change that should be available for CMS 4 projects as well.

lozcalver:
@GuySartorelli `shuffle()` was [added in 4.5.0](https://github.com/silverstripe/silverstripe-framework/commit/350888bf502ef1b79693b0e794745ea3593fdd71) (I only looked it up because I had no idea the method existed!) - if we were to merge this into the 3.12 branch, would we want to bump the framework requirement? Or should we just keep it simple and merge into 4?

GuySartorelli:
Ohhhh. That makes it a bit of a pain. I guess we could do feature detection and check if the method exists but given there was no info provided about the deprecated functionality this is avoiding, let's keep it simple and leave it targeting 4. We can always have a new pr for 3.12 if it becomes a problem 
for someone.
Thanks for spotting that @kinglozzer

GuySartorelli (approved):
This is fine as is for the branch tis targeting.

PR #732, opened 2023-11-16
remove plus icon for consistency

lerni:
Blog summary has for unknown reason a different icon than all other accordions. This PR removes the plus icon.

before:
before

after:
after


michalkleiner (approved):
No idea about the historical reason either, but consistency is good.

PR #739, opened 2024-01-11
ENH Add generic types

GuySartorelli:
Requires https://github.com/silverstripe/silverstripe-framework/pull/11108 and the associated PRs in silverstripe/cms and silverstripe/lumberjack for these generics to take effect.

## Issue
- https://github.com/silverstripe/silverstripe-framework/issues/11066

### DIFF
--- a/src/Model/BlogController.php
+++ b/src/Model/BlogController.php
@@ -15,7 +15,11 @@ use SilverStripe\Blog\Model\BlogTag;
 use SilverStripe\Blog\Model\BlogCategory;
 use SilverStripe\View\Parsers\URLSegmentFilter;
 use SilverStripe\Control\HTTPRequest;
+use SilverStripe\ORM\SS_List;
 
+/**
+ * @extends PageController<Blog>
+ */
 class BlogController extends PageController
 {
     /**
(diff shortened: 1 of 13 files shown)


src/Model/BlogController.php
```diff
@@ -443,7 +443,7 @@ public function getFilterDescription()
     /**
      * Returns a list of paginated blog posts based on the BlogPost dataList.
      *
-     * @return PaginatedList
+     * @return PaginatedList<ArrayList, BlogPost>
```
emteknetnz: Could you just double check this is correct - `$allPosts = $this->blogPosts ?: ArrayList::create();` -  `$this->blogPosts` has a docblock var of DataList and when it's assigned by methods like getCurrentProfilePosts() it's a DataList

emteknetnz (changes requested):

src/Model/BlogController.php
ChloeSartorelli: Ahh, I must have missed that, good call. I'll just use `SS_List` instead - this method will probably mostly be called in a template anyway so doesn't matter too much.

emteknetnz (approved):

Issue #748, opened 2024-03-14
Member BlogProfileImage doesn't get published unless FirstName/Surname is changed


purplespider:
### Module version(s) affected

4.1.2

### Description

Due to the order in `onBeforeWrite`, the `BlogProfileImage` added in `BlogMemberExtension` doesn't get automatically published unless you have also changed the member's name (unlikely).

Is there a reason that `$owns` isn't used to ensure the image gets published here?

### How to reproduce

1. Create a new user in Security.
2. Click **Create**
3. **Upload** a new "Blog profile image" for the user.
4. Click **Save**

**What I expect to happen:** The image gets published.
**What actually happens:** The image stays as "Draft".

### Validations

- [X] Check that there isn't already an issue that reports the same bug
- [X] Double check that your reproduction steps work in a fresh installation of [`silverstripe/installer`](https://github.com/silverstripe/silverstripe-installer) (with any code examples you've provided)

## PRs
- https://github.com/silverstripe/silverstripe-blog/pull/749
- #750

PR #750, opened 2024-03-19
Changed BlogProfileImage publish from onBeforeWrite to $owns

purplespider:
_Redo of #749_
## Description

Due to the order in `onBeforeWrite`, the `BlogProfileImage` added in `BlogMemberExtension` doesn't get automatically published unless you have also changed the member's name (unlikely).

This is because if `FirstName` or `Surname` are unchanged, then `onBeforeWrite` returns, before being able to publish the profile image.

This PR sets it to use `$owns` to take care of publishing the image instead. 

## Manual testing steps

1. Create a new user in Security.
2. Click **Create**
3. **Upload** a new "Blog profile image" for the user.
4. Click **Save**

**What I expect to happen:** The image gets published.
**What actually happens:** The image stays as "Draft".

## Issues

- #748

## Pull request checklist

- [x] The target branch is correct
    - See [picking the right version](https://docs.silverstripe.org/en/contributing/code/#picking-the-right-version)
- [x] All commits are relevant to the purpose of the PR (e.g. no debug statements, unrelated refactoring, or arbitrary linting)
    - Small amounts of additional linting are usually okay, but if it makes it hard to concentrate on the relevant changes, ask for the unrelated changes to be reverted, and submitted as a separate PR.
- [x] The commit messages follow our [commit message guidelines](https://docs.silverstripe.org/en/contributing/code/#commit-messages)
- [x] The PR follows our [contribution guidelines](https://docs.silverstripe.org/en/contributing/code/)
- [x] Code changes follow our [coding conventions](https://docs.silverstripe.org/en/contributing/coding_conventions/)
- [x] This change is covered with tests (or tests aren't necessary for this change)
- [x] Any relevant User Help/Developer documentation is updated; for impactful changes, information is added to the changelog for the intended release
- [x] CI is green

### DIFF
--- a/src/Model/BlogMemberExtension.php
+++ b/src/Model/BlogMemberExtension.php
@@ -35,6 +35,10 @@ class BlogMemberExtension extends DataExtension
         'BlogProfileImage' => Image::class
     ];
 
+    private static array $owns = [
+        'BlogProfileImage',
+    ];
+
     /**
      * @var array
      */
@@ -58,11 +62,6 @@ class BlogMemberExtension extends DataExtension
         while (!$this->validURLSegment()) {
             $this->owner->URLSegment = preg_replace('/-[0-9]+$/', '', $this->owner->URLSegment ?? '') . '-' . $count;
             $count++;
-        }
-
-        // Auto publish profile images
-        if ($this->owner->BlogProfileImage() && $this->owner->BlogProfileImage()->exists()) {
-            $this->owner->BlogProfileImage()->publishSingle();
         }
     }
 


michalkleiner (approved):

src/Model/BlogMemberExtension.php
```diff
@@ -35,6 +35,13 @@ class BlogMemberExtension extends DataExtension
         'BlogProfileImage' => Image::class
     ];
 
+    /**
+     * @var array
+     */
+    private static $owns = [
```
ChloeSartorelli:
```suggestion
    private static array $owns = [
```
No need to use PHPDoc when we're declaring new properties - we can just directly use typehints.

For now the typehint doesn't actually enforce anything, but in the future the config might actually validate types against typehints, so we try to add them in whenever we're adding new config.

GuySartorelli (changes requested):
Great! Thanks for doing this. Just one tiny change.

src/Model/BlogMemberExtension.php
purplespider: Ah, good to know. Thanks. I was just making sure it was consistent with how it was done in `BlogPost.php`. https://github.com/silverstripe/silverstripe-blog/blob/37d2ccff76d04d6fd8c1f73146350a5e074d2652/src/Model/BlogPost.php#L88

michalkleiner: Thanks @purplespider! Yeah, there's sometimes a conflict between how it's done currently vs how it would be done ideally going forward.

I reckon in this case Guy could've just pushed the change, leave a comment explaining the same as above, and already have it merged ;-)

michalkleiner (approved):

michalkleiner:
I've updated the commit message to have the [type prefix](https://docs.silverstripe.org/en/5/contributing/code/#commit-messages) as well (it's a FIX) and for it to be a bit more descriptive.

src/Model/BlogMemberExtension.php
ChloeSartorelli: I could have done. There's no hard and fast rules about what to let through and what not to. On the one hand it doesn't ultimately affect anything *right now* but on the other it's a small change so easy enough to make before merging.
I erred on the side of asking for the change in this case 🤷‍♀️

PR #762, opened 2024-06-05
MNT Fix scss lint warnings

GuySartorelli:
## Issue
- https://github.com/silverstripe/.github/issues/228


client/src/styles/cms.scss
```diff
@@ -2,11 +2,14 @@
...
-#FeaturedImage .middleColumn {
-  clear: none;
-  float: left;
+.cms table.ss-gridfield-table tr td.MergeAction a {
+  display: block;
+  height: 100%;
+  width: 100%;
 }
 
+/* stylelint-disable max-nesting-depth */
```
ChloeSartorelli, line 11: This lint rule is really intended to stop us getting into this situation in the first place - resolving this retroactively would require changing the specificity of the selectors used for styling which is not in scope for this card.

emteknetnz (approved):

PR #763, opened 2024-06-07
MNT Remove tests for unsupported optional modules

emteknetnz:
Issue https://github.com/silverstripe/.github/issues/272

These are tests for optional modules that were supported in CMS 4, though are not supported in CMS 5

The behat test will cause the blog module erronously fail if from recipe-kitchen-sink as the widgets + comments modules are only in require-dev, and thus won't be installed under kitchen sink, though they are when installing from blog


GuySartorelli (approved):
Seems okay. If anyone in the community is still using those modules and we accidentally break things from this module, they can raise PRs to fix it.

PR #769, opened 2024-06-19
ENH Auto-scaffold blog categories and tags formfields

GuySartorelli:
See post-implementation note in the linked issue about these - short version is that this mirrors the implementation used in `BlogPost`, but I can't update `BlogPost` to take advantage of it just yet. That'll be handled in a separate issue.

## Issue
- https://github.com/silverstripe/.github/issues/262

### DIFF
--- a/src/Model/BlogCategory.php
+++ b/src/Model/BlogCategory.php
@@ -2,7 +2,10 @@
 
 namespace SilverStripe\Blog\Model;
 
+use SilverStripe\CMS\Model\SiteTree;
+use SilverStripe\Forms\FormField;
 use SilverStripe\ORM\DataObject;
+use SilverStripe\TagField\TagField;
 
 /**
  * A blog category for generalising blog posts.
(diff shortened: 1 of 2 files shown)


GuySartorelli:
JS CI failure is unrelated to these changes

src/Model/BlogCategory.php
```diff
@@ -55,6 +58,44 @@ class BlogCategory extends DataObject implements CategorisationObject
...
+        return $this->scaffoldFormFieldForManyRelation($relationName, $fieldTitle, $ownerRecord);
+    }
+
+    private function scaffoldFormFieldForManyRelation(
+        string $relationName,
+        ?string $fieldTitle,
+        DataObject $ownerRecord
+    ): FormField {
+        $parent = ($ownerRecord instanceof SiteTree) ? $ownerRecord->Parent() : null;
+        $field = TagField::create(
```
emteknetnz, line 87: Is there any reason to use TagField over SearchableMultiDropdownField?  From memory their UI is quite similar, though TagField could create (update?) new records from within the field (add a string, press enter, create/update a record), while SearchableMultiDropdownField is for linking existing records which I think is what we're doing here

emteknetnz (changes requested):

src/Model/BlogCategory.php
ChloeSartorelli: TagField is what it's already using so I'm just copying that logic across.
But yes, the main difference is the ability to create new tags/categories directly in the formfield itself, which makes it quicker to add them.

I'm not aiming to change the way people add or create tags or categories with this PR.

emteknetnz (approved):
Tested locally, looks good

PR #788, opened 2024-09-26
Fix blog plural en.yml

3Dgoo:
The English plural names for blog and blog post are incorrect. This fixes it up.


GuySartorelli (approved):
LGTM, I've also updated these in transifex, which I think takes precedence even for the source strings.

GuySartorelli:
CI failures are unrelated

PR #799, opened 2024-12-08
API Deprecate API that's changing in CMS 6

GuySartorelli:
## Issue
- https://github.com/silverstripe/silverstripe-cms/issues/2947

### DIFF
--- a/src/Model/Blog.php
+++ b/src/Model/Blog.php
@@ -131,12 +131,15 @@ class Blog extends Page implements PermissionProvider
 
     /**
      * @var string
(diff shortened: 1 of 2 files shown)


src/Model/Blog.php
```diff
@@ -131,12 +131,15 @@ class Blog extends Page implements PermissionProvider
 
     /**
      * @var string
-     * @deprecated 5.4.0 use class_description instead.
+     * @deprecated 4.4.0 use class_description instead.
```
ChloeSartorelli, line 134: Fixing version number

emteknetnz (approved):

PR #832, opened 2025-11-13
DOC Add composer install command and repository link

emteknetnz:
Issue https://github.com/silverstripe/.github/issues/465

### DIFF
--- a/docs/en/index.md
+++ b/docs/en/index.md
@@ -1,5 +1,15 @@
 # Blog documentation
 
+## Installation
+
+```bash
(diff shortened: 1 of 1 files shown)


GuySartorelli:
This isn't in the optional features section of the docs - is the intention to add it there for CMS 5?

emteknetnz:
Not the intention, just being consistent with other modules

GuySartorelli (approved):
Isn't needed but not worth discussing, doesn't do any harm, will just merge.

